PR #103, opened 2020-03-19
null_check pattern

### LINKED ISSUE #89: null_check pattern
This is the pattern:

```
class Foo {
  private String z;
  void x() {
    if (this.z == null) { // here!
      throw new RuntimeException("oops");
    }
  }
}
```

The check for `null` is the pattern.

The pattern should not check constructors.

tribals:
Add `null_check` pattern. Closes #89

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -157,3 +157,24 @@ public void start() {
 ```
 
 
+***
+
+*Title*: Null checks
+
+*Code*: **null_check**
+
+*Description*: If we check that something equals null (except in constructor)
+it is considered a pattern.
+
+*Examples*:
+
+```java
+class Foo {
+  private String z;
+  void x() {
+    if (this.z == null) { // here!
+      throw new RuntimeException("oops");
+    }
+  }
+}
+```
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -178,9 +178,9 @@ def main():
             }
 
             sorted_result = predict(input_params)
-            print('The contribution list of each patterns correspondingly for file {}:'.format(java_file))
             found_pattern = False
             code_lines = None
+            value = None
             for iter, (key, val) in enumerate(sorted_result.items()):
                 if key in patterns_list:
                     if not found_pattern:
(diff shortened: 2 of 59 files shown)


lyriccoder:
1) @yegor256  Should we check this patterns also (they are not catched by the provided code):

`boolean i = z == null;`
`boolean i = z != null;`
`luckyName != null ? luckyName : "No lucky name found";`
`if (z != null)`
`luckyName == null ? luckyName : "No lucky name found";`
`assert param != null;`
`assertThrows(NullPointerException.class, () -> wrapperSum(null, 2));`
`Optional<String> op2 
            = Optional.ofNullable(null); `
`if (one == null || two == null || three == null)`
`Objects.requireNonNull(bar, "bar must not be null");`
`public void accept(String param){
    if (null != param && !param.isEmpty())
        System.out.println(param);
}`

2) Could you please also check the line numbers in test cases, not a length of an array. 
E.g., `Pattern.value('java.java') == [3, 105, 660]`

tribals:
> 1. Could you please also check the line numbers in test cases, not a length of an array.
>    E.g., `Pattern.value('java.java') == [3, 105, 660]`

Good idea! Will do so.

tribals:
Actually it will catch any binary expression where the operation is "equal" and right operand is `null', be it in `if` condition (either single or multiple combined with Boolean operators), in assignment or ternary.

It _will not_ catch "not equal" operations, though. As well when `null` is the left operand - this is weird but valid case.

yegor256:
@lyriccoder looks good to me as is. We can later improve it.

tribals:
@lyriccoder @acheshkov Please, do review. I will merge PR then.

tribals:
@lyriccoder @yegor256 @acheshkov Can anyone please re-trigger Travis build in some way?

yegor256:
@rultor merge

yegor256:
@tribals tests fail here, sorry

yegor256:
@rultor merge

0crat:
Job `gh:yegor256/aibolit#103` is not assigned, can't get performer

PR #105, opened 2020-03-20
Added checking ambiguous in if-return statements

Vitaly-Protasov:
As we discussed, It was added an opportunity to detect if-else if statements with returns inside


yegor256:
@rultor merge

0crat:
Job `gh:yegor256/aibolit#105` is not assigned, can't get performer

Issue #116, opened 2020-03-26
Check Null: Improvement


lyriccoder:
We need to check the following cases also for `null_check` pattern:

1. boolean i = z == null;
2. boolean i = z != null;
3. luckyName != null ? luckyName : "No lucky name found";
4. if (z != null)
5. luckyName == null ? luckyName : "No lucky name found";
6. assert param != null;
7. assertThrows(NullPointerException.class, () -> wrapperSum(null, 2));
8. Optional<String> op2 = Optional.ofNullable(null);
9. if (one == null || two == null || three == null)
10. Objects.requireNonNull(bar, "bar must not be null");
11. public void accept(String param){ if (null != param && !param.isEmpty()) System.out.println(param); }

0crat:
@acheshkov[/z](https://www.0crat.com/u/acheshkov) please, pay attention to this issue

acheshkov:
@0crat in

acheshkov:
@0crat assign @tribals

0crat:
> @0crat assign @tribals  ([here](https://github.com/yegor256/aibolit/issues/116#issuecomment-604308563))

@acheshkov The job [#116](https://github.com/yegor256/aibolit/issues/116) assigned to @tribals[/z](https://www.0crat.com/u/tribals), here is [why](https://www.0crat.com/footprint/CT5UB7XAP/4d1776a2-00d6-4826-bf3c-cc1b4e5fbcd5); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0crat:
> @0crat assign @tribals  ([here](https://github.com/yegor256/aibolit/issues/116#issuecomment-604308563))

@acheshkov Job [#116](https://github.com/yegor256/aibolit/issues/116) is not in the agenda of @tribals[/z](https://www.0crat.com/u/tribals), can't set title

0crat:
[...]

Issue #130, opened 2020-03-27
Bring static type checking for whole project


tribals:
Python is dynamically typed language. But there is [ongoing work](https://docs.python.org/3/library/typing.html) for bringing static type checking to it.

The static type checking consists of two parts: type hints and third-party static type checkers (linters).

Let's take a closer look to both parts.

Type hints, well, is just hints. They are simple annotations of source code which explain what type of variables, function parameters and return values is expected. Python interpreter by itself simply *ignores it at all*.

Type checkers are third-party tools which do actual processing of these hints, and produce a result - is these types fulfilled or not. A couple of examples: [mypy](http://mypy-lang.org/), [pyright](https://github.com/microsoft/pyright).

So, in order to benefit from static typing you must do the both. Annotating sources with type hints is worthless at it's own. Even worse, without checking these annotations can be misleading.

So, the decision here is either "annotate and check" or "don't even bother with it".

0crat:
@acheshkov[/z](https://www.0crat.com/u/acheshkov) please, pay attention to this issue

acheshkov:
@tribals I agree

acheshkov:
@0crat in

acheshkov:
@0crat assign @tribals

0crat:
> @0crat assign @tribals  ([here](https://github.com/yegor256/aibolit/issues/130#issuecomment-607297750))

@acheshkov The job [#130](https://github.com/yegor256/aibolit/issues/130) assigned to @tribals[/z](https://www.0crat.com/u/tribals), here is [why](https://www.0crat.com/footprint/CT5UB7XAP/c779793e-541f-49a3-b09a-150d0bc7d5a3); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0crat:
[...]

PR #156, opened 2020-04-02
WIP: Null Check improvements

tribals:
For #116.


tribals:
In fact, it is very annoying to write such kind of tests. These can be written in another way:

```python
from unittest import TestCase

from javalang.parser import Parser
from javalang.tokenizer import tokenize

from aibolit.patterns.null_check.null_check import NullCheck


class TestNullCheck(TestCase):
    def test_null_check_comparison_result_assignment(self):I
        snippet = 'boolean i = z == null;'

        node = Parser(tokenize(snippet)).parse_block_statement()

        self.assertEqual(NullCheck()._traverse_node(node), [1])
```

Here is two questions:
1. Is such style of tests accepted? Or it is better to always write a "real" Java class?
2. It requires us to call private method. Is it accepted as well?

0crat:
This pull request [#156](https://github.com/yegor256/aibolit/issues/156) is assigned to @fanifieiev[/z](https://www.0crat.com/u/fanifieiev), here is [why](https://www.0crat.com/footprint/CT5UB7XAP/b20ab37f-7c0a-4949-953b-a743a9ad34ea); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @acheshkov[/z](https://www.0crat.com/u/acheshkov) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be a monetary reward for this job

acheshkov:
@lyriccoder @yegor256 please say yours two cents.

acheshkov:
The point makes sense. But I'm not sure that we can do the same for any pattern. Actually we can, but some patterns deal with the structure that span more than one line and hence you will be  going to put this code as text to your test case.   


```python
snippet = 'here will be multi line snippet;'
```

tribals:
This is completely not a problem since Python has multi-line strings:

```python
snippet = """\
public void MyMethod(int x) {
  if (x == null) {}
}
"""
```

The only thing need to be done is calling right method of `javalang.parser.Parser` - in this example, it must be `.parse_member_declaration()`.

In fact, I propose to write *every* Java example code, even complete classes, inside test itself. These examples seem to be small, so there is no problem in keeping them inside test.

tribals:
Since nobody reacted, I assume this style is accepted.

acheshkov:
Let's give it a shot and see

fanifieiev:
@tribals @acheshkov I would actually suggest having `value ` method that would accept 
`javalang.ast.Node` but not `filename`, so that there is no need to call `NullCheck()._traverse_node(node)`

`_traverse_node(node)` looks like a private/internal method.

So, I would suggest:
```
class NullCheck(object):
    def value(self, javalang.ast.Node):
       ....
```
So, if someone needs to get the source code from a filename:
`result = NullCheck().value(AST(filename))`

If someone need to get the source code from snippet:
`result = NullCheck().value(_parser(snippet).parse_block_statement())`
I would also suggest having a separate class:
```
class BlockStatement:
    def __init__(self, snippet):
        self.snippet = snippet

    def value():
        return _parser(snippet).parse_block_statement()
```

and use the above as:
`result = NullCheck().value(
        BlockStatement(
             snippet
        ).value()
 )`

tribals:
> @tribals @acheshkov I would actually suggest having `value ` method that would accept
> `javalang.ast.Node` but not `filename`, so that there is no need to call `NullCheck()._traverse_node(node)`

I'm absolutely agree with you! But this change deserves it's own issue. I would suggest to adapt to static type checking first, please see: #130 and #165. After doing so, we can introduce a real interface for patterns, rather an ephemeral one (also know as "Duck type"). The next step would be to enable type-check in CI pipeline, so no new patterns could be accepted without conforming to that `Pattern` interface.

fanifieiev:
> > @tribals @acheshkov I would actually suggest having `value ` method that would accept
> > `javalang.ast.Node` but not `filename`, so that there is no need to call `NullCheck()._traverse_node(node)`
> 
> I'm absolutely agree with you! But this change deserves it's own issue. I would suggest to adapt to static type checking first, please see: #130 and #165. After doing so, we can introduce a real interface for patterns, rather than an ephemeral one (also know as "Duck type"). The next step would be to enable type-check in CI pipeline, so no new patterns could be accepted without conforming to that `Pattern` interface.

@tribals Actually I created an issue long ago to create interface 
https://github.com/yegor256/aibolit/issues/123

tribals:
@fanifieiev Excellent! Haven't seen it, though.

Well, seems like everything already in place in order to make patterns more solid at it's behavior. We need only to close issues. :slightly_smiling_face:

0crat:
@fanifieiev[/z](https://www.0crat.com/u/fanifieiev) this job was assigned to you 5days ago. It will be taken away from you soon, unless you close it, see [§8](http://www.zerocracy.com/policy.html#8). Read [this](http://www.yegor256.com/2014/04/13/no-obligations-principle.html) and [this](http://www.yegor256.com/2014/11/24/principles-of-bug-tracking.html), please.

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
-        file = Path(self.cur_file_dir, 'NullCheck.java')
+        snippet = """\
+        if (this.z == null) { // here!
+            throw new RuntimeException("oops");
+        }
+        """
 
-        self.assertEqual(self.testClass.value(file), [5])
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_block_statement()), [1]
```
fanifieiev, line 43: @tribals I think it is better to:
`NullCheck()._traverse_node(_parser(snippet).parse_block_statement()), [1]` as each test method is a dedicated test unit.
It is also not good to have this  `_traverse_node` method calling here.

PS: Actually each pattern should accept the `snippet `or `filename `or etc via constructor, but not the `value `method, but this is out of scope of this PR.

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
+        snippet = """\
+        public NullCheck(String z) {
+            if (z == null) { // here!
+                throw new RuntimeException("oops");
+            }
+        }
+        """
+
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_member_declaration()),
```
fanifieiev, line 56: @tribals Same as above

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
+
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_block_statement()), [1]
+        )
+
+    def test_null_check_ternary(self):
+        snippet = 'luckyName == null ? luckyName : "No lucky name found";'
+
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_block_statement()), [1]
```
fanifieiev, line 71: @tribals Same as above

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
+
+    def test_null_check_not_equal_comparison(self):
+        snippet = """\
+        if (this.z != null) { // here!
+            throw new RuntimeException("oops");
+        }
+        """
+
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_block_statement()), [1]
```
fanifieiev, line 82: @tribals Same as above

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_member_declaration()),
+            [],
+        )
+
+    def test_null_check_comparison_result_assignment(self):
+        snippet = "boolean i = z == null;"
+
+        self.assertEqual(
+            self.pattern._traverse_node(_parser(snippet).parse_block_statement()), [1]
```
fanifieiev, line 64: @tribals Same as above

test/patterns/null_check/test_null_check.py
```diff
@@ -20,23 +20,68 @@
...
+from javalang.tokenizer import tokenize
+
 from aibolit.patterns.null_check.null_check import NullCheck
 
 
 class TestNullCheck(TestCase):
-    cur_file_dir = Path(os.path.realpath(__file__)).parent
-    testClass = NullCheck()
+    def setUp(self):
+        self.pattern = NullCheck()
```
fanifieiev, line 33: @tribals According to below comment, it is better to remove this part

test/patterns/null_check/test_null_check.py
tribals: There is no need to create `NullCheck` explicitly in each test method since `setUp` is called once per every test method. Please, see: https://docs.python.org/3/library/unittest.html#organizing-tests.

I agree, though, that it is not clear from source code itself.

> It is also not good to have this `_traverse_node` method calling here.

I absolutely agree. A method which name is prefixed with underscore considered a private one, and is not allowed to call directly. But, since there is only one possible way to pass already parsed AST to pattern now, I suggest to leave it as is until we will fix all patterns.

tribals:
@fanifieiev Please, take a look at issue itself: #116. There was some pattern which are not implemented yet in this PR. Namely:

```
assertThrows(NullPointerException.class, () -> wrapperSum(null, 2));
Optional op2 = Optional.ofNullable(null);
Objects.requireNonNull(bar, "bar must not be null");
```

Unfortunatelly, I don't understand, what exactly should be considered a pattern? I requested clarification from reporter, but there is still none.

Could we separate them to individual issue and accept work done for now?

fanifieiev:
> @fanifieiev Please, take a look at issue itself: #116. There was some patterns which are not implemented yet in this PR. Namely:
> 
> ```
> assertThrows(NullPointerException.class, () -> wrapperSum(null, 2));
> Optional op2 = Optional.ofNullable(null);
> Objects.requireNonNull(bar, "bar must not be null");
> ```
> 
> Unfortunatelly, I don't understand, what exactly should be considered a pattern? I requested clarification from reporter, but there is still none.
> 
> Could we separate them to individual issue and accept work done for now? It already covers the wast majority of null checks.

@tribals Absolutely, you can leave a `@todo` with the explanation of what to do. Please find an example of Todo [here](https://github.com/yegor256/aibolit/blob/a1c8969d7fa897ba8f263675eb89cb1dbb21365a/aibolit/patterns/var_middle/var_middle.py#L72) if you do not know how to create a todo

test/patterns/null_check/test_null_check.py
fanifieiev: > There is no need to create `NullCheck` explicitly in each test method since `setUp` is called once per every test method. Thus, `self.pattern` is a new one in each test. Please, see: https://docs.python.org/3/library/unittest.html#organizing-tests.
> 
> I agree, though, that it is not clear from source code itself.
> 
> > It is also not good to have this `_traverse_node` method calling here.
> 
> I absolutely agree. A method which name is prefixed with underscore considered a private one, and is not allowed to call directly. But, since there is only one possible way to pass already parsed AST to pattern now, I suggest to leave it as is until we will fix all patterns.

@tribals I know that `setup` does it for each test method, but I personally(not only me) find that setup method is very confusing and brings more confusions. 
I think it is better to create pattern object in each test method  separately.
@acheshkov WDYT ?

tribals: Well, personally I find the same about setup methods, too. So, it is simple to me to follow you advice.

tribals:
@fanifieiev Please, take a look.

fanifieiev:
@tribals Please create a @todo for this case https://github.com/yegor256/aibolit/pull/156#issuecomment-610612129

tribals:
@fanifieiev Done

acheshkov:
@rultor merge

PR #392, opened 2020-06-04
Traversing AST for var in the middle pattern.

aravij:
Var in the middle pattern is rewritten to traverse full tree instead of flattening it.
New test from the issue is added.

### DIFF
--- a/aibolit/patterns/if_return_if_detection/if_detection.py
+++ b/aibolit/patterns/if_return_if_detection/if_detection.py
@@ -1,5 +1,5 @@
 import javalang
-from aibolit.patterns.var_middle.var_middle import JavalangImproved
+from aibolit.utils.java_parser import JavalangImproved
 
 
 class newJavalangImproved(JavalangImproved):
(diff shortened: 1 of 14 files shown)


aravij:
1. Variable declarations are now allowed after lists initialization with curly braces. According test is added.
2. Class JavalangImproved is moved to separate module in utils and marked as deprecated. Patterns better traverse AST manually instead of flattening it. Lines class marked as deprecated too, because it is only usage is in JavalangImproved class.
3. New class Scope_status added to help traversing AST.
4. The pattern itself has changed. Not it allows calling constructor of super class before variable declarations. It is done, because such calls must be done first. Another change is allowing declaring variables after method calls, if such calls used to initialize other variables. Both this changes affected 'test_case_nested_class' test.
5. New test with super class method call is added.
6. Several test cases were named wrongly, the same as 'VarMiddleTest'. They have been renamed.

acheshkov:
@rultor merge

0crat:
Job `gh:yegor256/aibolit#392` is not assigned, can't get performer

PR #426, opened 2020-06-19
CFG. Classes design and simple implementation.

acheshkov:
I've outlined the methods I need to to able to build CFG. Simple implementation is added.


aibolit/utils/cfg_builder.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# from functools import lru_cache
```
aravij, line 23: Remove comments

aibolit/utils/cfg_builder.py
```diff
@@ -0,0 +1,61 @@
...
+
+NODE_TYPES = [
+    ASTNodeType.ASSIGNMENT,
+    ASTNodeType.RETURN_STATEMENT
+]
+
+
+def build_cfg(tree: AST) -> Graph:
+    '''Create Control Flow Graph'''
+    g = Graph()
```
aravij, line 38: You are using undirected graph. Is it meats you plan?

aibolit/utils/java_class_method.py
```diff
@@ -49,3 +50,8 @@ def name(self) -> str:
     @property
     def java_class(self) -> 'JavaClass':
         return self._java_class
+
+    def cfg(self) -> Graph:
```
aravij, line 54: It might be a property, because we are not intended to change AST anyhow.
Also just noticed, we could use single `@cached_property` decorator from `functools` instead of composing `@property` decorator with `@lru_cache`. It will assume that cache is used only for one property.

test/utils/test_cfg_builder.py
```diff
@@ -0,0 +1,38 @@
...
+from aibolit.utils.java_package import JavaPackage
+from aibolit.utils.java_class import JavaClass
+from aibolit.utils.java_class_method import JavaClassMethod
+
+
+class CFGBuilderTestCase(TestCase):
+
+    def test_cfg_of_method(self):
+        java_package = JavaPackage(Path(__file__).parent.absolute() / "SimpleClass.java")
+        fst: JavaClass = [c for c in java_package.java_classes][0]
```
aravij, line 35: We could omit type declarations. If types are not deduced automatically now, because of `@property  # type: ignore`, it will probably work after we would start using `@cached_property`. It seems to be supported by mypy.

acheshkov:
@rultor merge

0crat:
Job `gh:yegor256/aibolit#426` is not assigned, can't get performer

PR #474, opened 2020-07-01
non_final_class with AST

angusev:


aibolit/patterns/non_final_class/non_final_class.py
```diff
@@ -11,9 +11,15 @@ def __init__(self):
...
-            node.position.line for _, node in classes
-            if len([v for v in ['final', 'abstract'] if v in node.modifiers]) == 0
-        ]
+        tree = AST.build_from_javalang(build_ast(filename))
+        nodes = tree.nodes_by_type(ASTNodeType.CLASS_DECLARATION)
+
+        positions = []
+        for node in nodes:
+            line = tree.get_attr(node, 'source_code_line')
+            modifiers = tree.get_attr(node, 'modifiers')
```
aravij, line 20: There is no such thing as `modifiers` yet. All those fields will be added soon, but now you have to extract them on your own. If you print AST, you will see that CLASS_DECLARATION node has child with type COLLECTION. There are some nodes with strings, that will represent modifiers. To access child with particular type, you could use `children_with_type` method of AST.

aibolit/patterns/non_final_class/non_final_class.py
```diff
@@ -11,9 +11,15 @@ def __init__(self):
...
+
+            if len([v for v in ['final', 'abstract'] if v in modifiers]) == 0:
```
aravij, line 22: You better just intersect two sets and check its length.

lyriccoder, line 22: We should consider only main class, not inner classes. Is it right? In your example we count all inner classes

acheshkov:
@rultor merge

0crat:
Job `gh:yegor256/aibolit#474` is not assigned, can't get performer

PR #483, opened 2020-07-06
Filter test and train

lyriccoder:
fix #482

### DIFF
--- a/scripts/02-filter-and-move.py
+++ b/scripts/02-filter-and-move.py
@@ -58,7 +58,6 @@ parser.add_argument(
 args = parser.parse_args()
 MAX_CLASSES = args.max_classes
 TXT_OUT = 'found-java-files.txt'
-SPLIT_CSV = 'split.csv'
 CSV_OUT = '02-java-files.csv'
 
 DIR_TO_CREATE = 'target/02'
--- a/scripts/02-filter-and-move.py
+++ b/scripts/02-filter-and-move.py
@@ -48,9 +48,17 @@ parser.add_argument(
     required=False,
     default=None
 )
+parser.add_argument(
+    '--split_only',
+    required=False,
+    help='Only split filenames into train and test, do not filter',
+    default=False,
+    action='store_true'
+)
 args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


scripts/02-filter-and-move.py
```diff
@@ -196,17 +203,26 @@ def call_back():
...
+
+    if not args.split_only:
+        start = time.time()
+        results = walk_in_parallel()
+
+        if not os.path.isdir(DIR_TO_CREATE):
+            os.makedirs(DIR_TO_CREATE)
+
+        df = pd.DataFrame(results, columns=['filename', 'class_type'])
+        df = df[df['class_type'] == 999]
```
aravij, line 217: Can we give a name to this magical constant?

scripts/02-filter-and-move.py
```diff
@@ -196,17 +203,26 @@ def call_back():
...
+        df.to_csv(path_csv_out, index=False)
+        df['filename'].to_csv(path_txt_out, header=None, index=None)
+        end = time.time()
+        print('It took ' + str(end - start) + ' seconds')
+    from sklearn.model_selection import train_test_split
```
aravij, line 222: Why do we import at the end of file?

scripts/02-filter-and-move.py
```diff
@@ -196,17 +203,26 @@ def call_back():
...
     path_csv_out = str(Path(current_location, DIR_TO_CREATE, CSV_OUT))
     path_txt_out = str(Path(current_location, DIR_TO_CREATE, TXT_OUT))
-    df = pd.DataFrame(results, columns=['filename', 'class_type'])
-    df = df[df['class_type'] == 999]
-    df.to_csv(path_csv_out, index=False)
-    df['filename'].to_csv(path_txt_out, header=None, index=None)
-    end = time.time()
-    print('It took ' + str(end - start) + ' seconds')
+
+    if not args.split_only:
```
aravij, line 209: Can you clarify the logic behind this `if`.
If there was no command line argument `--split-only`, then we do what?

PR #492, opened 2020-07-06
Refactored Supermethod pattern

Vitaly-Protasov:


test/patterns/supermethod/test_supermethod.py
```diff
@@ -37,7 +37,7 @@ def test_empty(self):
     def test_instance_of(self):
         # It has 2 matches in anonymous class!
         file = str(Path(self.cur_file_dir, 'Anonymous.java'))
-        self.assertEqual(len(self.testClass.value(file)), 2)
+        self.assertEqual(len(self.testClass.value(file)), 1)
```
aravij, line 40: Please explain why do we change tests?

aibolit/patterns/supermethod/supermethod.py
```diff
@@ -20,27 +35,16 @@ def value(self, filename: str):
...
         source_code = read_text_with_autodetected_encoding(filename)
-        tree = parse(source_code)
         lines = source_code.splitlines()
-        for _, method_decl_node in tree.filter(MethodDeclaration):
-            code_line = method_decl_node.position.line
-            for _, super_method_inv in method_decl_node.filter(SuperMethodInvocation):
+        tree = JavaPackage(filename)
+        for method_decl_node in tree.nodes_by_type(ASTNodeType.METHOD_DECLARATION):
+            code_line = tree.get_attr(method_decl_node, 'source_code_line')
+            for super_method_inv in tree.nodes_by_type(ASTNodeType.SUPER_METHOD_INVOCATION):
```
aravij, line 42: Is it something with me, but always when I see `inv` I think about some sort of inverse. And by the way, you cannot declare super method, it is already declared in class parent, so, basically, the only thing you can do with super method is to invoke them.

aibolit/patterns/supermethod/supermethod.py
```diff
@@ -1,15 +1,30 @@
...
+            # Checking not to start with '/' is aimed to get
+            # rid of comments, which are all childs of node.
+            # We check the occurance any letter in name in order
+            # to get rid of '' string and None.
+            if not method_name.startswith('/') and re.search(r'[^\W\d]', method_name) is not None:
+                extracted_name = method_name
+                return extracted_name
+        return ''
+
+    def value(self, filename: str) -> Set[int]:
```
aravij, line 27: Don't we usually return lists in patterns?

aibolit/patterns/supermethod/supermethod.py
```diff
@@ -20,27 +35,16 @@ def value(self, filename: str):
         """
         results = []
         source_code = read_text_with_autodetected_encoding(filename)
-        tree = parse(source_code)
         lines = source_code.splitlines()
-        for _, method_decl_node in tree.filter(MethodDeclaration):
-            code_line = method_decl_node.position.line
-            for _, super_method_inv in method_decl_node.filter(SuperMethodInvocation):
+        tree = JavaPackage(filename)
+        for method_decl_node in tree.nodes_by_type(ASTNodeType.METHOD_DECLARATION):
```
aravij, line 40: It sounds to me, that outer loop is redundant and just brings repetition of code lines, which was tried to removed by using set. If you iterate over all super method invocations, isn't it enough to find all its lines? By the way, there are probably will be no source code available for patterns in the future. We shouldn't stick to using it.

aravij:
I accidentally approved it. There is o way to request changes without adding comments. Please resolve comments before merging.

test/patterns/supermethod/test_supermethod.py
Vitaly-Protasov: Test are changed, because, for the previous version of this pattern, as you noticed, code brings repetition of code lines. That is why I used set in order to resolve it

aibolit/patterns/supermethod/supermethod.py
Vitaly-Protasov: yeap, agree

aibolit/patterns/supermethod/supermethod.py
Vitaly-Protasov: Hah, I decided not to chan original names:)

aibolit/patterns/supermethod/supermethod.py
Vitaly-Protasov: defenitely agree. Fixed it in the last commit

acheshkov:
@rultor merge

aravij:
@acheshkov please merge

acheshkov:
@rultor merge

Issue #462, opened 2020-06-29
SuppressWarnings annotation support


g4s8:
JDK has source annotation to suppress any warnings: `@SupressWarnings`: https://docs.oracle.com/javase/8/docs/api/java/lang/SuppressWarnings.html Many linters (like javac or PMD) suppores this annotation to ignore warning, e.g.
```java
@SuppressWarnings("PMD.ProhibitPublicStaticMethods")
public static void foo() { }
```
In java code will suppress public static method PMD warning. It's the standard way to ignore warnings for Java linters and checkers.

I'd suggest supporting this annotation for Aibolit to suppress warnings for classes methods and variables, it may use "aibolit" prefix to separate from other suppression, like:
```java
@SuppressWarnings("aibolit.P22")
public void foo(byte[] array) { }
```

0crat:
@acheshkov[/z](https://www.0crat.com/u/acheshkov) please, pay attention to this issue

yegor256:
@acheshkov @g4s8 definitely a feature we need!

0crat:
I'm not managing `cqfn/aibolit` GitHub repository, you have to contact me in Slack first, see [§11](http://www.zerocracy.com/policy.html#11)

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 4@ceb75b72-5128-4a32-a6a0-db018bc785eb, thread: pool-409-thread-1
com.zerocracy.radars.github.ThrottledComments[74] java.lang.IllegalStateException: Can't post an identical message to cqfn/aibolit#462: I'm not managing `cqfn/aibolit` GitHub repository, you have to contact me in Slack first, see [§11](
```

[0.53.15](https://github.com/zerocracy/farm): Issue: cqfn/aibolit#462, Action: `closed`

PR #494, opened 2020-07-07
Function for find start and end lines for node

lyriccoder:
Fixing https://github.com/cqfn/aibolit/issues/495
part of #462

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -46,6 +46,8 @@ from aibolit import __version__
 from aibolit.config import Config
 from aibolit.ml_pipeline.ml_pipeline import train_process, collect_dataset
 from aibolit.model.model import TwoFoldRankingModel, Dataset  # type: ignore  # noqa: F401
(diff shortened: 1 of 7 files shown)


Vitaly-Protasov:
Think it can be rewritten to the new interface:
```
from aibolit.ast_framework import AST, ASTNodeType
from aibolit.ast_framework.java_package import JavaPackage


def flatten(l):
    return [item for sublist in l for item in sublist]


def find_start_and_end_lines(node):
    ast = JavaPackage(filename)
    max_line = ast.get_attr(node, 'source_code_line')

    def traverse(node):
        for child in ast.tree.succ[node]:
            child_line = ast.get_attr(child, 'source_code_line')
            if child_line is not None:
                nonlocal max_line
                if child_line > max_line:
                    max_line = child_line

    traverse(node)
    return node.position.line, max_line
    
   ```

acheshkov:
@rultor merge

PR #496, opened 2020-07-07
Add javalang fields as networkx attributes

aravij:
javalang fields are added as `networkx` attributes. If field was a javalang `Node` or list of `Nodes`, it is now a named tuple `ASTNodeReference` with node index (int).
`ast_node_type.py` file now has map with attributes names for each `Node` types.
Static methods of `AST`, which were building `AST` from `javalang` have been refactored to be more consistent.
Method `get_subtrees` now return `AST` instead of list of nodes indexes.
Property `nodes` removed. Now you should use `get_nodes` method to get all nodes. To filter nodes with type, provide this type as method argument.
All other files, except `ast.py` and `ast_noe_type.py` has changes, only because of API change.


0crat:
I'm not managing `cqfn/aibolit` GitHub repository, you have to contact me in Slack first, see [§11](http://www.zerocracy.com/policy.html#11)

aibolit/ast_framework/ast.py
```diff
@@ -67,7 +71,7 @@ def __str__(self) -> str:
                 depth -= print_step
         return printed_graph
 
-    def subtrees_with_root_type(self, root_type: ASTNodeType) -> Iterator[List[int]]:
+    def get_subtrees(self, root_type: ASTNodeType) -> Iterator['AST']:
```
acheshkov, line 74:
```suggestion
    def get_subtrees(self, root_type: ASTNodeType) -> Iterator[AST]:
```

can we use type itself?

aravij: No, we are in the middle of `AST` class. It is not defined yet. If we would use simply `Iterator[AST]`, then we would get an error, something like "undefined token". To prevent it, we should wrap types in quotes. We should do it only in a case class `A` use itself in its methods as argument or return type.

aravij:
> Too large (

I am trying to make it smaller and smaller over the time. Hope we reach satisfied size for us both.

aibolit/ast_framework/ast.py
```diff
@@ -187,56 +187,96 @@ def get_binary_operation_params(self, binary_operation_node: int) -> BinaryOpera
         return BinaryOperationParams(self.get_attr(operation_node, 'string'), left_side_node, right_side_node)
 
     @staticmethod
-    def _build_from_javalang(tree: DiGraph, javalang_node: Node) -> int:
-        node_index = len(tree) + 1
-        tree.add_node(node_index)
-        AST._extract_javalang_node_attributes(tree, javalang_node, node_index)
-        AST._iterate_over_children_list(tree, javalang_node.children, node_index)
+    def _add_subtree_from_javalang_node(tree: DiGraph, javalang_node: Union[Node, Set[Any], str],
+                                        javalang_node_to_index_map: Dict[Node, int]) -> int:
```
acheshkov, line 191:
```suggestion
    def _add_subtree_from_javalang_node(
      tree: DiGraph, 
        avalang_node: Union[Node, Set[Any], str],
        javalang_node_to_index_map: Dict[Node, int]
     ) -> int:
```

acheshkov:
@rultor merge

PR #504, opened 2020-07-09
Ignore patterns with suppress_warnings

### LINKED ISSUE #462: SuppressWarnings annotation support

lyriccoder:
fix #462

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -48,6 +48,7 @@ from pkg_resources import parse_version
 from aibolit import __version__
 from aibolit.config import Config
 from aibolit.ml_pipeline.ml_pipeline import train_process, collect_dataset
+from aibolit.utils.ast_builder import build_ast
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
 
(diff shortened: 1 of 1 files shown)


lyriccoder:
Here is the file I tested the feature:
[LottieImageAsset.txt](https://github.com/cqfn/aibolit/files/4895847/LottieImageAsset.txt)

If I run aibolit by the `aibolit check --filename ...\LottieImageAsset.java --full` command, will get the output:

```
D:\target\0001\temp\LottieImageAsset.java score: 7.7
D:\target\0001\temp\LottieImageAsset.java[58]: Count If Return (P6: 4.90 1/4)
D:\target\0001\temp\LottieImageAsset.java[58]: Null check (P13: 1.40 2/4)
D:\target\0001\temp\LottieImageAsset.java[21]: Non final attribute (P12: 0.80 3/4)
D:\target\0001\temp\LottieImageAsset.java[11]: Non final class (P24: 0.60 4/4)
Total score: 7.70, files seen: 1, ncss 2
```

1) If you replace the @SuppressWarnings({"aibolit.P23", "aibolit.P11"}) warning with @SuppressWarnings({"aibolit.P23", "aibolit.P24"}) at the 10th line, you won't see `P24` pattern (suppress for the whole class declaration)

2) If you replace 56th line with   @SuppressWarnings({"aibolit.P6", "aibolit.P13"}), you will ignore `P6, P13` patterns (for the whole function declaration)

3) If you replace 20th line with     @SuppressWarnings({"aibolit.P23", "aibolit.P12"}), you will ignore `P12` pattern (for the field declaration)

So, 3 types annotations are supported: for a class, for a function and for a field

aibolit/__main__.py
```diff
@@ -327,6 +328,23 @@ def run_recommend_for_file(file: str, args):
...
+    patterns_ignored = defaultdict(list)
+
+    for node, patterns_list in functions_with_annotations.items():
+        start_pos, end_pos = find_start_and_end_lines(node)
+        for p in patterns_list:
+            patterns_ignored[p].append([start_pos, end_pos])
+
+    for node, patterns_list in fields_with_annotations.items():
+        for p in patterns_list:
+            patterns_ignored[p].append([node.position.line, node.position.line])
```
aravij, line 346: Why duplicates?

lyriccoder: because `field declaration`  is  usually 1 line. And we do not find start and end position for it.

acheshkov:
@rultor merge

acheshkov:
@rultor release, tag=`1.2.3rc1`

Issue #520, opened 2020-07-15
Show progress for intergration test.


aravij:
Integration tests get slower and Travis kill its for not responding for 10 minutes. Printing progress should solve this problem.
Tasks:
 - [x] Print and update a single line progress, showing quantity of files processed out of total number of files
 - [x] Make exception messages more informative by adding information about exception it caused

PR #523, opened 2020-07-16
Progress bar in testing + exceptions

Vitaly-Protasov:
Solved #520


test/integration/all.py
```diff
@@ -1,28 +1,32 @@
...
-        except Exception:
+        except Exception as e:
             print(
                 "Error apply the pattern:",
                 pattern['name'],
                 pattern['code'],
                 "to file",
                 filename
             )
+            print(e)
```
aravij, line 29: We better make a singe `print` statement with one formatting, maybe multi line string. So it could be read as a single sentence.
Same in another `except`

Vitaly-Protasov: done

aravij:
@acheshkov please merge

Issue #528, opened 2020-07-20
Refactor 18 patterns


Vitaly-Protasov:
Description of each pattern you can find here: [PATTERNS.md](https://github.com/cqfn/aibolit/blob/master/PATTERNS.md)
- [x] 1) many_primary_ctors
- [x] 2) method_chaining
- [x] 3) method_siblings
- [x] 4) multiply_try
- [x] 5) nested_blocks
- [x] 6) nested_loop
- [x] 7) non_final_attribute
- [x] 8) null_check
- [x] 9) partial_synchronized
- [x] 10) private_static_method
- [x] 11) protected_method
- [x] 12) public_static_method
- [x] 13) redundant_catch
- [x] 14) send_null
- [x] 15) string_concat
- [x] 16) this_finder
- [ ] 17) var_decl_diff
- [x] 18) var_siblings

aravij:
I think last patterns should be distributed the next way:
 - @Vitaly-Protasov can take `nesting_block` and `nesting_loop` (last one should be deleted, check my comment in #562).
 - I am currently working with grouping statements by scope and providing `JavaLangImproved` functionality. I consider to take patterns, which use it:
    - `partial_synchronized`
    - `redundant_catch`
    - `var_decl_diff`
 - In order to spread the knowledge of using new `ast_framework`, @lyriccoder could take the rest of the patterns:
    - `send_null`
    - `this_finder`

This will finish our epic patterns refactoring (but not metrics!).

Vitaly-Protasov:
> I think last patterns should be distributed the next way:
> 
> * @Vitaly-Protasov can take `nesting_block` and `nesting_loop` (last one should be deleted, check my comment in #562).
> * I am currently working with grouping statements by scope and providing `JavaLangImproved` functionality. I consider to take patterns, which use it:
>   
>   * `partial_synchronized`
>   * `redundant_catch`
>   * `var_decl_diff`
> * In order to spread the knowledge of using new `ast_framework`, @lyriccoder could take the rest of the patterns:
>   
>   * `send_null`
>   * `this_finder`
> 
> This will finish our epic patterns refactoring (but not metrics!).

I have almost finished "redundant_catch". may i take this?

aravij:
@Vitaly-Protasov Yep, no problem.

PR #542, opened 2020-07-20
refactored method_siblings

Vitaly-Protasov:
In case of #528

### DIFF
--- a/aibolit/patterns/method_siblings/method_siblings.py
+++ b/aibolit/patterns/method_siblings/method_siblings.py
@@ -32,8 +32,8 @@ class MethodSiblings:
     start with the same names
     '''
     def _is_method_names_close(self, node: ASTNode, new_node: ASTNode) -> bool:
-        node_name = re.split('([A-Z][^A-Z]*)', node.name)
-        return len(node_name) > 1 and new_node.name.startswith(node_name[0])
+        splited_name = re.split('([A-Z][^A-Z]*)', node.name)
+        return len(splited_name) > 1 and new_node.name.startswith(splited_name[0])
 
     def value(self, filename: str) -> List[int]:
         numbers: List[int] = []
--- a/aibolit/patterns/method_siblings/method_siblings.py
+++ b/aibolit/patterns/method_siblings/method_siblings.py
@@ -33,8 +33,7 @@ class MethodSiblings:
     '''
     def _is_method_names_close(self, node: ASTNode, new_node: ASTNode) -> bool:
         node_name = re.split('([A-Z][^A-Z]*)', node.name)
-        return node.node_index < new_node.node_index and \
-            len(node_name) > 1 and new_node.name.startswith(node_name[0])
+        return len(node_name) > 1 and new_node.name.startswith(node_name[0])
 
     def value(self, filename: str) -> List[int]:
         numbers: List[int] = []
@@ -42,6 +41,6 @@ class MethodSiblings:
 
         for node in method_nodes:
             for new_node in method_nodes:
-                if self._is_method_names_close(node, new_node):
+                if node.node_index < new_node.node_index and self._is_method_names_close(node, new_node):
                     numbers.extend([node.line, new_node.line])
         return sorted(list(set(numbers)))
--- a/aibolit/patterns/method_siblings/method_siblings.py
+++ b/aibolit/patterns/method_siblings/method_siblings.py
@@ -37,7 +37,8 @@ class MethodSiblings:
 
     def value(self, filename: str) -> List[int]:
         numbers: List[int] = []
-        method_nodes = AST.build_from_javalang(build_ast(filename)).get_proxy_nodes(ASTNodeType.METHOD_DECLARATION)
+        ast = AST.build_from_javalang(build_ast(filename))
+        method_nodes = ast.get_proxy_nodes(ASTNodeType.METHOD_DECLARATION)
 
         for node in method_nodes:
             for new_node in method_nodes:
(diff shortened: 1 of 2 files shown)


aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,30 @@
...
 class MethodSiblings:
-    def __init__(self):
-        pass
-
-    def value(self, filename: str) -> List[LineNumber]:
+    '''
+    Count those methods, which
+    start with the same names
+    '''
+    def __compare_next_nodes(self, node: ASTNode, new_node: ASTNode) -> List[int]:
```
aravij, line 34: - Seems like we do not compare **next nodes**, but we compare **methods names**. You can mention this both in method name in method arguments.
 - The result of compression is usually `bool`, lines could be added to list in `value` method.

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,30 @@
...
+        node_name = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', node.name)
```
aravij, line 35: We better simplify regex here.

aravij: And `node_name` is more likely be `common_names_prefix`. Isn't it?

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,30 @@
...
+        if node.node_index < new_node.node_index and len(node_name) > 1 and \
+           new_node.name.startswith(node_name[0]):
+            return [node.line, new_node.line]
+        return []
```
aravij, line 39: If we will return simply `bool`, we could just return expression.

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,30 @@
...
-            composed = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', node.name)
-            if len(composed) > 1:
-                for _, inner in build_ast(filename).filter(javalang.tree.MethodDeclaration):
-                    if node.name != inner.name:
-                        if inner.name.startswith(composed[0]):
-                            numbers.append(node.position.line)
-        return numbers
+    def value(self, filename: str) -> List[int]:
+        numbers: List[int] = []
+        method_nodes = AST.build_from_javalang(
```
aravij, line 43: we better declare `ast` on separate line.

aibolit/patterns/method_siblings/method_siblings.py
aravij: Maybe even better name is `_is_method_names_close`.
Also I think for private methods _one_ underscore is enough.

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,30 @@
...
-        pass
-
-    def value(self, filename: str) -> List[LineNumber]:
+    '''
+    Count those methods, which
+    start with the same names
+    '''
+    def __compare_next_nodes(self, node: ASTNode, new_node: ASTNode) -> List[int]:
+        node_name = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', node.name)
+        if node.node_index < new_node.node_index and len(node_name) > 1 and \
```
aravij, line 36: We better move this compassion `node.node_index < new_node.node_index` to `value` method.

aibolit/patterns/method_siblings/method_siblings.py
Vitaly-Protasov: done

aibolit/patterns/method_siblings/method_siblings.py
Vitaly-Protasov: simplified

nope, it is not the the common_names_prefix. here i have the full name

aibolit/patterns/method_siblings/method_siblings.py
Vitaly-Protasov: done!

aibolit/patterns/method_siblings/method_siblings.py
Vitaly-Protasov: done!

aibolit/patterns/method_siblings/method_siblings.py
Vitaly-Protasov: Agree with you

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,28 @@
...
+    start with the same names
+    '''
+    def _is_method_names_close(self, node: ASTNode, new_node: ASTNode) -> bool:
+        node_name = re.split('([A-Z][^A-Z]*)', node.name)
+        return len(node_name) > 1 and new_node.name.startswith(node_name[0])
 
-    def value(self, filename: str) -> List[LineNumber]:
+    def value(self, filename: str) -> List[int]:
+        numbers: List[int] = []
+        method_nodes = AST.build_from_javalang(build_ast(filename)).get_proxy_nodes(ASTNodeType.METHOD_DECLARATION)
```
aravij, line 40: We better move `ast` creation to separate line.

aibolit/patterns/method_siblings/method_siblings.py
```diff
@@ -19,26 +19,28 @@
...
 
 class MethodSiblings:
-    def __init__(self):
-        pass
+    '''
+    Count those methods, which
+    start with the same names
+    '''
+    def _is_method_names_close(self, node: ASTNode, new_node: ASTNode) -> bool:
+        node_name = re.split('([A-Z][^A-Z]*)', node.name)
```
aravij, line 35: It is better use `[0]` right here, it will allow to give better name to `node_name`. Maybe something like this
```suggestion
        common_prefix = re.split('([A-Z][^A-Z]*)', node.name)[0]
```
And change the next line accordingly.

Vitaly-Protasov: unfortunately i cant change it, because, after that I check `len(node_name) > 1`

aravij: Then we better pick up a different name.

acheshkov:
@rultor merge

PR #543, opened 2020-07-20
Parsing nested fields for replacing javalang nodes.

aravij:
Test for complex fields is added.


acheshkov:
@aravij  what is a "complex field"?

aravij:
@acheshkov There are some fields, that are represented by list, which has other lists as its items and so on. Here I iterate over such lists recursively and replace `ASTNodeReference` objects with `ASTNode` objects.

acheshkov:
@rultor merge

PR #545, opened 2020-07-20
refactored NonFinalAttribute

Vitaly-Protasov:
Solved in case of #528

### DIFF
--- a/aibolit/patterns/non_final_attribute/non_final_attribute.py
+++ b/aibolit/patterns/non_final_attribute/non_final_attribute.py
@@ -1,21 +1,43 @@
-
-from typing import List
-
-from javalang.tree import FieldDeclaration, ClassDeclaration
-
-from aibolit.types_decl import LineNumber
+# The MIT License (MIT)
+#
+# Copyright (c) 2020 Aibolit
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


aibolit/patterns/non_final_attribute/non_final_attribute.py
```diff
@@ -1,21 +1,43 @@
...
 
 
 class NonFinalAttribute:
+    '''
+    return lines of non-final attributes
+    '''
+    def _check_final(self, field: ASTNode) -> bool:
+        if 'final' not in field.modifiers:
+            return True
+        return False
```
aravij, line 35:
```suggestion
        return 'final' not in field.modifiers
```

aibolit/patterns/non_final_attribute/non_final_attribute.py
```diff
@@ -1,21 +1,43 @@
...
 from aibolit.utils.ast_builder import build_ast
+from typing import List
+from aibolit.ast_framework.ast_node import ASTNode
 
 
 class NonFinalAttribute:
+    '''
+    return lines of non-final attributes
+    '''
+    def _check_final(self, field: ASTNode) -> bool:
```
aravij, line 32: If we will make this function one liner, we may just substitute its code right into `value` function.

aibolit/patterns/non_final_attribute/non_final_attribute.py
Vitaly-Protasov: got it

aibolit/patterns/non_final_attribute/non_final_attribute.py
Vitaly-Protasov: done

acheshkov:
@rultor merge

PR #553, opened 2020-07-21
refactored var_siblings pattern

Vitaly-Protasov:
Solved in case of #528


aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+                local_var_node).get_proxy_nodes(ASTNodeType.VARIABLE_DECLARATOR))[0]
+            var_name = var_declaration.name
+            if var_name not in vars_info:
+                # to filter not complex names
+                temp_name = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', var_name)
+                if len(temp_name) > 1:
+                    vars_info[var_name] = var_line
+        return vars_info
+
+    def _compare_nodes_names(self, node1_name: str, node2_name: str) -> bool:
```
aravij, line 48: Does `node_name` is actually `var_name`? Both in method name and in arguments. And I think numbers better be used at the end of variable names.

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        # here we want to find those names, which start with the same word
```
aravij, line 49: here we `check` not `find`.

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        # and the lenght of this word > 3
+        node_name = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', node1_name)
```
aravij, line 51: - We could move expression `re.findall(...)` to a separate method with name explaining the regex
 - We already have here `node1_name` and `node2_name`. Another `node_name` variable seem like better be renamed. Is it `common_names_prefix` or something like this?

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        if node2_name != node1_name and \
```
aravij, line 52: I think you here trying to exclude the situation with comparing name to itself. It better be moved out into `_find_sibling_vars` method.

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+                node2_name.startswith(node_name[0]) and len(node_name[0]) > 3:
+            return True
+        return False
```
aravij, line 55: We better return  result of some expression, than check it with `if`.
Instead of
```python
if some_function():
    return True
return False
```
use
```python
return some_function()
```
or, if necessary
```python
return bool(some_function())
```

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+
+    def _find_sibling_vars(self, vars_info: Dict) -> List[int]:
```
aravij, line 57: Provide full type with `Dict`.

aravij: May be specify what king of info in `vars_info`. Something like `vars_lines`.

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        lines: List[int] = []
+        for node in vars_info.keys():
+            for new_node in vars_info.keys():
```
aravij, line 60:
```suggestion
        for node in vars_info:
            for new_node in vars_info:
```

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
 
 
 class VarSiblings:
-    def __init__(self):
-        pass
+    '''
+    Find those variables, which have complex
+    names and start with the same word
+    '''
+    def _traverse_method_vars(self, ast: AST, node: ASTNode) -> Dict:
```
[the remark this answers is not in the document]
Vitaly-Protasov: fixed

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        vars_info: Dict = {}
```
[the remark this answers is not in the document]
Vitaly-Protasov: added

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: 1)
2) yes, you are right

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: I dont think that it is a good idea, especially for comparing name I made the function _compare_nodes_names
As for me, it looks more obvious to deal with this action here.

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: okay, got it

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: agree

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: fixed

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+        for local_var_node in ast.get_proxy_nodes(ASTNodeType.LOCAL_VARIABLE_DECLARATION):
+            var_line = local_var_node.line
+            var_declaration = list(ast.get_subtree(
+                local_var_node).get_proxy_nodes(ASTNodeType.VARIABLE_DECLARATOR))[0]
+            var_name = var_declaration.name
```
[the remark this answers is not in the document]
Vitaly-Protasov: done!

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,71 @@
...
+            if var_name not in vars_info:
+                # to filter not complex names
+                temp_name = re.findall(r'[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)', var_name)
```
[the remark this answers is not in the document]
Vitaly-Protasov: rewrote

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,69 @@
...
+    '''
+    Find those variables, which have complex
+    names and start with the same word
+    '''
+    def _collect_method_variables_names(self, ast: AST, node: ASTNode) -> Dict[str, int]:
+        assert node.node_type == ASTNodeType.METHOD_DECLARATION
+        vars_lines: Dict[str, int] = {}
+        for local_var_node in ast.get_proxy_nodes(ASTNodeType.LOCAL_VARIABLE_DECLARATION):
+            var_line = local_var_node.line
+            var_declaration = local_var_node.declarators[0]
```
aravij, line 39: There might be declared several fields in single line. You could add them all to `var_lines`.

aibolit/patterns/var_siblings/var_siblings.py
aravij: As you said, your function `_compare_nodes_names` should comparing nodes. Your comment in function says that if two names are identical, they are close. So it is strange to see it in such function.
If you try to avoid comparing variable name with itself, you better do it outside of this function. Otherwise your function trying to complete two tasks: compare names  and avoid same names. Functions should complete a singe tasks.

aibolit/patterns/var_siblings/var_siblings.py
```diff
@@ -1,26 +1,69 @@
...
+            var_name = var_declaration.name
+            if var_name not in vars_lines:
+                # to filter not complex names
+                temp_name = re.split('([A-Z][^A-Z]*)', var_name)
+                if len(temp_name) > 1:
+                    vars_lines[var_name] = var_line
+        return vars_lines
+
+    def _compare_nodes_names(self, node_name_1: str, node_name_2: str) -> bool:
```
aravij, line 48: We might use here same name as in previous PR `_is_names_close`

aibolit/patterns/var_siblings/var_siblings.py
aravij: Seems like you forget to commit changes.

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: ok

aibolit/patterns/var_siblings/var_siblings.py
Vitaly-Protasov: done

acheshkov:
@rultor merge

PR #555, opened 2020-07-21
Check return objects of metrics and patterns in integration test.

aravij:


aravij:
Source code line number algorithm is now embeded in `ASTNode`. If there is no line field at givven node, it takes all node, reachable from givven and take minimum from line fields of all such nodes. If even this does not helps, we traverse from givven node to root and seek for line field. If this approach fails too, exception is raised.

acheshkov:
Great!

acheshkov:
@rultor merge

PR #558, opened 2020-07-22
Computed fields for ASTNode.

aravij:
Computed fields are fields which are computed based on fields that came from `javalang`.
There are provided several standard computed fields as:
 - `fields`, `methods`, `constructors` for class, interface and annotation a declarations
 - `name` field for class field and variable declarations

Adding manual computed properties can be done with `computed_fields_registry`.

All registered computed properties are available as any other usual property of `ASTNode`.

Misc:
 - Better printing multi line comments when printing `ASTNode`
 - Skipped test for java class decomposition. It uses deprecated API and got broken after a slight change in java source file.


aibolit/ast_framework/computed_fields_registry.py
```diff
@@ -0,0 +1,69 @@
...
+    def register(
+        self,
+        compute_field: Callable[["ASTNode"], Any],
+        name: str,
+        *node_types: "ASTNodeType",
+    ) -> None:
+        for node_type in node_types:
+            computed_fields = self._registry[node_type]
+            if name in computed_fields and \
+               not _ComputedFieldsRegistry._is_in_interactive_shell():
```
acheshkov, line 45: why do you check interactive mode?

aravij: If you are in interactive mode, rules are softer and you are allowed to redefine compured properties for easy prototyping. But if we launch aibolit as script or module, fields redefinition is forbidden, because it sounds like field name typo.

aibolit/ast_framework/computed_fields_catalog/chained_fields.py
```diff
@@ -0,0 +1,63 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from typing import Union, Any, Callable
+from itertools import chain
+
+from aibolit.ast_framework import ASTNode
+
+
+def chain_field_getter_factory(*steps: Union[str, int]) -> Callable[[ASTNode], Any]:
```
acheshkov, line 29: Do we have tests for that function?

aibolit/ast_framework/computed_fields_catalog/nodes_filter.py
```diff
@@ -0,0 +1,45 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from typing import Callable, Iterator
+
+from aibolit.ast_framework import ASTNode, ASTNodeType
+
+
+def nodes_filter_factory(
```
acheshkov, line 28: Could we add comments?

acheshkov, line 28: Do we have tests for that function?

aibolit/ast_framework/computed_fields_registry.py
```diff
@@ -0,0 +1,69 @@
...
+
+from collections import defaultdict
+import sys
+from typing import Dict, Callable, Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from aibolit.ast_framework import ASTNode, ASTNodeType  # noqa: F401
+
+
+class _ComputedFieldsRegistry:
```
acheshkov, line 31: Do we have tests for that class?

aibolit/ast_framework/computed_fields_catalog/chained_fields.py
aravij: In `test_ast_node.py` I test both `chain_field` and `nodes_filter` fields. They are checked as node fields, not as separate function, so it is just one test for both function and architecture to provide those functions as fields.

aibolit/ast_framework/computed_fields_catalog/nodes_filter.py
aravij: It is tested in the same test as `chain_field_factory` in `test_ast_node.py`.

aibolit/ast_framework/computed_fields_registry.py
aravij: No, this class is quite simple. It register functions in dictionary, provide read access to its item and may clear it. Nothing much. If you consider to test the function which checks if we are in interactive mode or not, that it is auxiliary functionality, it just simplifies prototyping. If it get broken and will affect production use case, we will catch it with `test_ast_node.py`. After all to test such functionality `unittest` package is not enough. We have to execute our code in different environment.

aravij:
> Could we add comments? 

Added.

aibolit/ast_framework/computed_fields_catalog/nodes_filter.py
aravij: Added.

acheshkov:
@rultor merge

PR #559, opened 2020-07-22
Correct NCSS metric comment and count try keyword.

aravij:


lyriccoder:
is there a test case when we do not have try statement? we should also consider case when we do not count it.

aravij:
> is there a test case when we do not have try statement? we should also consider case when we do not count it.

Yes we have them, for example here https://github.com/cqfn/aibolit/blob/master/test/metrics/ncss/ChainedIfElse.java

acheshkov:
@rultor merge

PR #560, opened 2020-07-23
refactored StringConcatFinder pattern

Vitaly-Protasov:
Solved in case of #528


aibolit/patterns/string_concat/string_concat.py
```diff
@@ -1,54 +1,50 @@
...
-                    if is_string_literal:
-                        if node.operandl.position:
-                            lines.add(node.operandl.position.line)
-                        elif node.operandr.position:
-                            lines.add(node.operandr.position.line)
-                        elif hasattr(node.operandr, '_position'):
-                            lines.add(node.operandr._position.line)
-                        elif hasattr(node.operandl, '_position'):
-                            lines.add(node.operandl._position.line)
+                if hasattr(node.operandl, 'value') and isinstance(node.operandl.value, str) and \
```
aravij, line 36: Can we, instead of checking for attribute existing, check for node type? Do we expect here only `ASTNodeType.LITERAL` or anything else?

aibolit/patterns/string_concat/string_concat.py
```diff
@@ -1,54 +1,50 @@
...
+                        not node.operandl.value.isdigit():
+                    if node.operandl.line:
```
aravij, line 38: Now property `line` automatically search for `line` among children, so you could just add `node.line`.

aibolit/patterns/string_concat/string_concat.py
```diff
@@ -1,54 +1,50 @@
...
+                        lines.add(node.operandl.line)
+                    elif node.operandr.line:
+                        lines.add(node.operandr.line)
+
+                elif hasattr(node.operandr, 'value') and isinstance(node.operandr.value, str) and \
```
aravij, line 43: It sounds like we check only if one argument of `BINARY_OPERATION` is string, while comment in class says we are looking for both arguments to be string. If that correct, we have to check both operands to be string.

test/patterns/string_concat/test_concat_string.py
```diff
@@ -63,7 +63,7 @@ def test_string_with_quotes(self):
         self.assertEqual(lines, [
             352, 372, 373, 379, 380, 408, 409, 418, 422, 429, 430, 438, 456, 466,
             479, 494, 499, 509, 562, 563, 604, 605, 610, 701, 995, 1009, 1012, 1029,
-            1032, 1080, 1258, 1267, 1337, 1386, 1414
+            1032, 1080, 1188, 1258, 1267, 1337, 1366, 1372, 1386, 1414
```
aravij, line 66: Can you explain why does test changed?

aibolit/patterns/string_concat/string_concat.py
Vitaly-Protasov: we expect 3 different types,  ofc we  can just check is it one of these or not, but I think that hasattr is better to

aibolit/patterns/string_concat/string_concat.py
Vitaly-Protasov: +

aibolit/patterns/string_concat/string_concat.py
lyriccoder: we search  for any string concatenation. but there can be different concatenation:
`"String" + "4"`
`int a = "b";`
`a+ "b";`
`b + "a";`
`"; " + message.toString()`
```

            Integer d = 1 + b;
            System.out.println(a + b);
```
etc.

look at the tests, there are examples there.

test/patterns/string_concat/test_concat_string.py
lyriccoder: it  was not found earlier:
`return "\"" + (String) p.getDefault() + "\".to_string()";`
`example = "\"" + ((param.example != null) ? param.example : "") + "\".to_string()";`
It's correct. Seems previous code didn't find double quotes

aibolit/patterns/string_concat/string_concat.py
aravij: So it just look for using concatenation with any string literal?
If so, we better create a single function to check for a string literal and use it to check left and right side of operators.

aibolit/patterns/string_concat/string_concat.py
aravij: Well than mention them in a comment at least, cause I just can't imagine what are the others two.

aibolit/patterns/string_concat/string_concat.py
lyriccoder: ok

aibolit/patterns/string_concat/string_concat.py
Vitaly-Protasov: done

aibolit/patterns/string_concat/string_concat.py
```diff
@@ -33,18 +33,20 @@ def value(self, filename: str) -> List[int]:
         ast = AST.build_from_javalang(build_ast(filename))
         for node in ast.get_proxy_nodes(ASTNodeType.BINARY_OPERATION):
             if node.operator == '+':
+                # by this check we want to filter nodes and consider only
+                # [Literal, MemberReference, MethodInvocation, This]
```
aravij, line 37: How did you get this list?
When I check node types for `value` field, I get following:
 - ASSERT_STATEMENT
 - ASSIGNMENT
 - ELEMENT_VALUE_PAIR
 - LITERAL
 - TRY_RESOURCE

By the way, can you explain, how other than LITERAL nodes meet this pattern?

aibolit/patterns/string_concat/string_concat.py
```diff
@@ -33,18 +33,20 @@ def value(self, filename: str) -> List[int]:
...
                 if hasattr(node.operandl, 'value') and isinstance(node.operandl.value, str) and \
                         not node.operandl.value.isdigit():
-                    if node.operandl.line:
-                        lines.add(node.operandl.line)
-                    elif node.operandr.line:
+                    if node.line:
+                        lines.add(node.line)
+                    elif node.line:
```
aravij, line 42: Please pay attention to the code you write.

aibolit/patterns/string_concat/string_concat.py
```diff
@@ -33,18 +33,20 @@ def value(self, filename: str) -> List[int]:
...
                 elif hasattr(node.operandr, 'value') and isinstance(node.operandr.value, str) and \
                         not node.operandr.value.isdigit():
-                    if node.operandl.line:
-                        lines.add(node.operandl.line)
-                    elif node.operandr.line:
-                        lines.add(node.operandr.line)
+                    if node.line:
+                        lines.add(node.line)
+                    elif node.line:
+                        lines.add(node.line)
```
aravij, line 50: Also here

aibolit/patterns/string_concat/string_concat.py
aravij: We expect `.line` to return always `int`. So simply add it to the set.

aibolit/patterns/string_concat/string_concat.py
aravij: @Vitaly-Protasov please move logic of checking operand to separate function and call it in `operandr` and `operandl` here.

acheshkov:
@rultor merge

PR #563, opened 2020-07-23
RFC metric.

aravij:
Calculating RFC metric as:
 - Count all **public** methods
 - Count distinct method invocations in **public** methods. Method invocation a considered distinct:
   - if they have different names
   - in case of same names, if one is local and another is not (called on some object, not on 'this')
 - Invocation of **inherited** methods do not count. Invocation considered to be of inherited method, if:
    - invocation is local, i.e. method is just invoked as opposite to be invoked on some object **and**
    - there is no such method name among all method names of current class


acheshkov:
@rultor merge

PR #564, opened 2020-07-23
FanOut metric.

aravij:
- Metric is simplified. All previous tests passed, except for `test_2`. We have been counting object names used in method invocation as types. This was fixed.
 - Currently we simply count all distinct REFERENCE_TYPE (excluding package names).
 - Tests added.


aibolit/metrics/fanout/FanOut.py
```diff
@@ -1,66 +1,112 @@
...
+    def value(self, ast: AST) -> int:
+        fan_out = 0
+        for class_declaration in ast.get_subtrees(ASTNodeType.CLASS_DECLARATION):
+            fan_out += self._calculate_class_fan_out(class_declaration)
+
+        return fan_out
+
+    def _calculate_class_fan_out(self, java_class: AST) -> int:
+        class_declaration = java_class.get_root()
+        assert class_declaration.node_type == ASTNodeType.CLASS_DECLARATION
```
Vitaly-Protasov, line 42: Don't you think that this assert is EXTRA? In function value in cycle FOR you iterate through each node with type `ASTNodeType.CLASS_DECLARATION`. After this node goes to the function `_calculate_class_fan_out`. What else can goes wrong?

aibolit/metrics/fanout/FanOut.py
```diff
@@ -1,66 +1,112 @@
...
+        return len(used_classes_names)
 
-        p = JavaPackage(filename)
-        for class_name in p.java_classes:
-            tree = p.java_classes[class_name]
-            for var_node in tree.get_nodes(ASTNodeType.VARIABLE_DECLARATOR):
-                var_child = list(tree.children_with_type(var_node, ASTNodeType.STRING))
-                new_class_name = tree.get_attr(var_child[0], 'string')
+    def _get_class_name_from_type_reference(self, type_reference: ASTNode) -> str:
+        assert type_reference.node_type == ASTNodeType.REFERENCE_TYPE
```
Vitaly-Protasov, line 56: Same this

aibolit/metrics/fanout/FanOut.py
aravij: In current version of code this `assert` will never raise an exception. But later on, the method, calling this one may change. I want to guarantee, that further code of this method may rely on some solid prerequisites. It is like additional type checking. Because of dynamic nature of `ASTNode` we can't guarantee, weather or not this field is available with mypy. But with such `asssert` we get this guarantee. It is a defense programming, when  your function trust no one and tries to check input parameters as hard as possible.

acheshkov:
@rultor merge

Issue #565, opened 2020-07-24
If file is failed, we should keep making processing `recommend`


lyriccoder:
If java parsers is failed, we should keep making processing `recommend`

lyriccoder:
https://github.com/artipie/http/blob/ac23b6bc3e9a63dbd1e6026112e9345cf5f2462a/src/main/java/com/artipie/http/stream/ByteByByteSplit.java

PR #569, opened 2020-07-24
Fix bug when aibolit fails if we have any error when pattern calculation

lyriccoder:
fix https://github.com/cqfn/aibolit/issues/565

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -259,7 +259,7 @@ def find_start_and_end_lines(node) -> Tuple[int, int]:  # noqa: C901
 
 def calculate_patterns_and_metrics(file, args):
     code_lines_dict = input_params = {}  # type: ignore
-    error_string = None
+    error_exc = None
     patterns_to_suppress = args.suppress
     try:
         config = Config.get_patterns_config()
@@ -276,10 +276,10 @@ def calculate_patterns_and_metrics(file, args):
                 continue
             __count_value(metric, input_params, code_lines_dict, file, is_metric=True)
     except Exception as ex:
-        error_string = str(ex)
+        error_exc = ex
         input_params = []  # type: ignore
 
-    return input_params, code_lines_dict, error_string
+    return input_params, code_lines_dict, error_exc
 
 
 def inference(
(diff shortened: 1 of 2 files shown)


aibolit/__main__.py
```diff
@@ -486,15 +492,21 @@ def get_exit_code(results):
...
     perfect_code_number = 0
+    errors_strings = []
     for result_for_file in results:
         results = result_for_file.get('results')
         errors_string = result_for_file.get('error_string')
         if not results and not errors_string:
             perfect_code_number += 1
         elif not results and errors_string:
-            errors_number += 1
+            if 'javalang.parser.JavaSyntaxError' not in errors_string:
```
acheshkov, line 502: Is it possible to match Exception by type not by strings?

aibolit/__main__.py
```diff
@@ -486,15 +492,21 @@ def get_exit_code(results):
...
+                errors_strings.append(errors_string)
+                errors_number += 1
+            else:
+                # ignore JavaSyntaxError, it is expected error
+                perfect_code_number +=1
 
-    if errors_number == files_analyzed:
+    if len(errors_strings) == files_analyzed:
         # we have errors everywhere
```
acheshkov, line 510: > # we have errors everywhere

what do you mean by "we have errors everywhere"?

aibolit/__main__.py
```diff
@@ -330,54 +330,60 @@ def inference(
 
     return results
 
-
-def run_recommend_for_file(file: str, args):
+# flake8: noqa: C901
```
acheshkov, line 333: can we not ignore it?

aibolit/__main__.py
lyriccoder: no, because we keep only string here

aibolit/__main__.py
lyriccoder: it means if we have 10 errors and we analyzed 10 files

aibolit/__main__.py
lyriccoder: after 1 if statement was added, the function began to show it. It is difficult to split it

test/recommend/test_recommend_pipeline.py
```diff
@@ -175,7 +177,7 @@ def test_text_format(self):
         new_mock = format_converter_for_pattern(mock_input)
         text = create_text(new_mock, full_report=True)
         md5_hash = md5('\n'.join(text).encode('utf-8'))
-        self.assertEqual(md5_hash.hexdigest(), '5efdc7930a66874e9d0c7dcfef734687')
+        self.assertEqual(md5_hash.hexdigest(), '2a67e22091ba2cfd76847d292ed90142')
```
acheshkov, line 180: why is hash changed?

aibolit/__main__.py
```diff
@@ -495,16 +509,16 @@ def get_exit_code(results):
...
-        errors_string = result_for_file.get('error_string')
-        if not results and not errors_string:
+        ex = result_for_file.get('exception')
+        if not results and not ex:
             perfect_code_number += 1
-        elif not results and errors_string:
-            if 'javalang.parser.JavaSyntaxError' not in errors_string:
-                errors_strings.append(errors_string)
+        elif not results and ex:
+            if not isinstance(ex, javalang.parser.JavaSyntaxError):
```
acheshkov, line 516: could you import just `JavaSyntaxError` ?

lyriccoder: ok but there is no difference. If we import it it won't become just JavaSyntaxError. Import is just an acronym. It will still be javalang.parser.JavaSyntaxError

test/recommend/test_recommend_pipeline.py
lyriccoder: Error occured was replaced with Error occurred

acheshkov:
@rultor merge

PR #573, opened 2020-07-28
Refactored partially synchronized methods pattern.

aravij:
- Test data is simplifyed
 - Old test data moved into integration test samples
 - After refactoring all old tests have passed
изображение

Most of the changes are due to files renaming and moving.


aibolit/ast_framework/ast.py
```diff
@@ -264,9 +264,20 @@ def _add_javalang_standard_node(tree: DiGraph, javalang_node: Node) -> Tuple[int
...
 
+        AST._post_process_javalang_attributes(tree, node_type, attributes)
+
         tree.add_node(node_index, **attributes)
         return node_index, node_type
 
+    @staticmethod
+    def _post_process_javalang_attributes(tree: DiGraph, node_type: ASTNodeType, attributes: Dict[str, Any]) -> None:
+        """
+        Replace some attributes with more appropriate values for convince work
```
acheshkov, line 275:
```suggestion
        Replace some attributes with more appropriate values for convenient work
```

aibolit/ast_framework/ast.py
```diff
@@ -80,7 +80,7 @@ def __iter__(self) -> Iterator[ASTNode]:
         for node_index in self.tree.nodes:
             yield ASTNode(self.tree, node_index)
 
-    def get_subtrees(self, root_type: ASTNodeType) -> Iterator['AST']:
+    def get_subtrees(self, *root_type: ASTNodeType) -> Iterator['AST']:
```
acheshkov, line 83: Does `root_type` has list type?

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def value(self, filename: str) -> List[int]:
+        ast = AST.build_from_javalang(build_ast(filename))
+        lines: List[int] = []
+        for method_ast in ast.get_subtrees(
+            ASTNodeType.METHOD_DECLARATION, ASTNodeType.CONSTRUCTOR_DECLARATION
+        ):
+            method_declaration = method_ast.get_root()
+            is_fully_sync_method = len(method_declaration.body) == 1 and \
```
acheshkov, line 43: > len(method_declaration.body) == 1

method declaration is a list and it may have zero length?

acheshkov: can it have length > 1 ?

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
```diff
@@ -0,0 +1,53 @@
...
+                method_declaration.body[0].node_type == ASTNodeType.SYNCHRONIZED_STATEMENT
```
acheshkov, line 44: > method_declaration.body[0]

here you check the type first child node of the body?

aibolit/ast_framework/ast.py
aravij: Yes, this is mypy syntax. Looks strange. You specify just single item type, but get list if them.

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
aravij: It appeared that yes. It may have zero length if method is abstract or native (implemented with c++). I also found some examples of code with just public methods with no body. I couldn't manage to figure out how is this possible, but we have examples of such code in integration tests.

To your second question, what do you mean? Can any method has several statements or fully synchronized method. If any method, then sure, it can have as many statements as it needs. But fully synchronized methods consist only of a single `synchronized` statement, with so,e statements nested in it. Notice, that `method_declaration.body` only contains top level statements, same as in `javalang`.

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
aravij: Exactly. If method consist of single statement and it is `synchronized`, than it is a fully synchronized method.

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
acheshkov: If `method_declaration.body` is either None or some value, then I beg a question: Why we need `list` type for `body` property?

aravij: There **were** some situations, where this field was `None`. I changed this behavior, so in such case we replace it with empty list. So it is always a list, but sometimes empty.

test/patterns/partially_synchronized_methods/FullSynchronization.java
```diff
@@ -0,0 +1,15 @@
...
+    int synchronizationField = 0;
+
+    public SingleClass() {
+        synchronized(synchronizationField) {
+            int x = 0;
+        }
+    }
+
+    public void method() {
+        synchronized(synchronizationField) {
```
lyriccoder, line 11: can you add in this case different type of synchronization? We had a lot of test when we had Class inside, array, function, etc.

aravij: There is no point in it. Current pattern do nothing with class, functions and nested scopes. So it really does not matter what is inside `synchronized` block, it all ignored.

aibolit/patterns/partially_synchronized_methods/partially_synchronized_methods.py
acheshkov: Why we need `list` type to store optional value?

aravij: @acheshkov  Because now you can iterate over the `body` nevertheless it is empty or not. If there is chances it is `None`, then you have to check it every time before `for` cycle. Now the value is consistent, and can always be iterated over.

acheshkov: clear

acheshkov:
@rultor merge

acheshkov:
@rultor merge

PR #576, opened 2020-07-28
Rewrite Send null with ASTNode

lyriccoder:
fix https://github.com/cqfn/aibolit/issues/575

### DIFF
--- a/aibolit/patterns/send_null/send_null.py
+++ b/aibolit/patterns/send_null/send_null.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Any
 
 from aibolit.ast_framework import ASTNodeType, AST
 from aibolit.utils.ast_builder import build_ast
@@ -9,19 +9,14 @@ class SendNull:
     def __init__(self):
         pass
 
-    def __is_null(self, val):
-        has_ternary_true_value = hasattr(val, 'value')
-        if has_ternary_true_value:
-            is_ternary_true_str = isinstance(val.value, str)
-            if is_ternary_true_str:
-                if val.value == 'null':
-                    return True
-                else:
-                    return False
-            else:
-                return False
-        else:
+    def __is_null(self, val: Any) -> bool:
+        if not hasattr(val, 'value'):
             return False
+        if not isinstance(val.value, str):
+            return False
+        if val.value != 'null':
+            return False
+        return True
 
     def value(self, filename: str) -> List[int]:
 
--- a/aibolit/patterns/send_null/send_null.py
+++ b/aibolit/patterns/send_null/send_null.py
@@ -1,8 +1,8 @@
 import javalang
 from typing import List
 
+from aibolit.ast_framework import ASTNodeType, AST
 from aibolit.utils.ast_builder import build_ast
-
 
 class SendNull:
 
@@ -65,3 +65,41 @@ class SendNull:
                 pass
         lst = sorted(lst)
         return lst
+
+
+class SendNullFixed:
+
+    def __init__(self):
+        pass
+
+
+    def value(self, filename: str) -> List[int]:
+
+        lines = set()
+        tree = AST.build_from_javalang(build_ast(filename))
+        invocatios_types = [
+            ASTNodeType.METHOD_INVOCATION,
+            ASTNodeType.EXPLICIT_CONSTRUCTOR_INVOCATION,
+            ASTNodeType.CLASS_CREATOR
+        ]
+        # print(tree)
+        for node in tree.get_proxy_nodes(*invocatios_types):
+            # print(node.member, '------------------------------')
+            # if node.line > 3533:
+                # print(node)
+            for argument in node.arguments:
+                if (argument.node_type == ASTNodeType.LITERAL) and (argument.value == "null"):
+                    lines.add(argument.line)
+
+        for node in tree.get_proxy_nodes(ASTNodeType.TERNARY_EXPRESSION):
+            # if node.line > 1000:
+                # print(node)
+            if (hasattr(node.if_true, 'value') and node.if_true.value == 'null') or \
+                    (hasattr(node.if_false, 'value') and node.if_false.value == 'null'):
+                lines.add(node.line)
+
+        lst = sorted(lines)
+        return lst
+
+# a = SendNullFixed().value(r'D:\git\aibolit\test\patterns\send_null\Temp.java')
+# print(a)
--- a/aibolit/patterns/send_null/send_null.py
+++ b/aibolit/patterns/send_null/send_null.py
@@ -24,8 +24,12 @@ class SendNull:
                     lines.add(argument.line)
 
         for node in tree.get_proxy_nodes(ASTNodeType.TERNARY_EXPRESSION):
-            if (hasattr(node.if_true, 'value') and node.if_true.value == 'null') or \
-                    (hasattr(node.if_false, 'value') and node.if_false.value == 'null'):
+            has_ternary_true_value = hasattr(node.if_true, 'value')
+            is_ternary_true_str = isinstance(node.if_true, str)
+            has_ternary_false_value = hasattr(node.if_false, 'value')
+            is_ternary_false_str = isinstance(node.if_false, str)
+            if (has_ternary_true_value and is_ternary_true_str and node.if_true.value == 'null') or \
+                    (has_ternary_false_value and is_ternary_false_str and node.if_false.value == 'null'):
                 lines.add(node.line)
 
         lst = sorted(lines)
--- a/aibolit/patterns/send_null/send_null.py
+++ b/aibolit/patterns/send_null/send_null.py
@@ -9,6 +9,20 @@ class SendNull:
     def __init__(self):
         pass
 
+    def __is_null(self, val):
+        has_ternary_true_value = hasattr(val, 'value')
+        if has_ternary_true_value:
+            is_ternary_true_str = isinstance(val.value, str)
+            if is_ternary_true_str:
+                if val.value == 'null':
+                    return True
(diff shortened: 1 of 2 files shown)


aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,28 @@ class SendNull:
...
+            ASTNodeType.EXPLICIT_CONSTRUCTOR_INVOCATION,
+            ASTNodeType.CLASS_CREATOR
+        ]
+        for node in tree.get_proxy_nodes(*invocatios_types):
+            for argument in node.arguments:
+                if (argument.node_type == ASTNodeType.LITERAL) and (argument.value == "null"):
+                    lines.add(argument.line)
+
+        for node in tree.get_proxy_nodes(ASTNodeType.TERNARY_EXPRESSION):
+            has_ternary_true_value = hasattr(node.if_true, 'value')
```
aravij, line 27: Can we check just node type, instead of using `hasattr`.

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,28 @@ class SendNull:
...
+            is_ternary_true_str = isinstance(node.if_true, str)
```
aravij, line 28: I consider literal field value to be string. Is there are some examples, when we get some other data?

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,28 @@ class SendNull:
...
+        lines = set()
+        tree = AST.build_from_javalang(build_ast(filename))
+        invocatios_types = [
+            ASTNodeType.METHOD_INVOCATION,
+            ASTNodeType.EXPLICIT_CONSTRUCTOR_INVOCATION,
+            ASTNodeType.CLASS_CREATOR
+        ]
+        for node in tree.get_proxy_nodes(*invocatios_types):
+            for argument in node.arguments:
+                if (argument.node_type == ASTNodeType.LITERAL) and (argument.value == "null"):
```
aravij, line 23: Maybe just omit parentheses?

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,28 @@ class SendNull:
...
+        for node in tree.get_proxy_nodes(ASTNodeType.TERNARY_EXPRESSION):
+            has_ternary_true_value = hasattr(node.if_true, 'value')
+            is_ternary_true_str = isinstance(node.if_true, str)
+            has_ternary_false_value = hasattr(node.if_false, 'value')
+            is_ternary_false_str = isinstance(node.if_false, str)
+            if (has_ternary_true_value and is_ternary_true_str and node.if_true.value == 'null') or \
+                    (has_ternary_false_value and is_ternary_false_str and node.if_false.value == 'null'):
+                lines.add(node.line)
+
+        lst = sorted(lines)
```
aravij, line 35: I think all lines will come up already sorted. So maybe use list for lines?

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,28 @@ class SendNull:
...
+            for argument in node.arguments:
+                if (argument.node_type == ASTNodeType.LITERAL) and (argument.value == "null"):
+                    lines.add(argument.line)
+
+        for node in tree.get_proxy_nodes(ASTNodeType.TERNARY_EXPRESSION):
+            has_ternary_true_value = hasattr(node.if_true, 'value')
+            is_ternary_true_str = isinstance(node.if_true, str)
+            has_ternary_false_value = hasattr(node.if_false, 'value')
+            is_ternary_false_str = isinstance(node.if_false, str)
+            if (has_ternary_true_value and is_ternary_true_str and node.if_true.value == 'null') or \
```
aravij, line 31: This long check might be moved out into a separate method, so we could check null for `if_false` and `if_true`.

test/patterns/send_null/test_send_null.py
```diff
@@ -14,7 +14,7 @@ def test_one_send(self):
 
     def test_multi_level_invocation(self):
         lines = self.method_send_null_finder.value(Path(self.dir_path, 'Configuration.java'))
-        self.assertEqual(lines, [379, 442, 549, 638, 656, 830, 866, 1362, 2393, 2988, 3080, 3492, 3855])
+        self.assertEqual(lines, [379, 442, 549, 638, 656, 830, 866, 1362, 2393, 2874, 2988, 3080, 3492, 3758, 3855])
```
aravij, line 17: Sounds like we found some new lines, that match pattern.

aibolit/patterns/send_null/send_null.py
lyriccoder: yep, sometimes it is a class

aibolit/patterns/send_null/send_null.py
lyriccoder: cannot understand, sometimes we do not have `value` field, we need to check it

aibolit/patterns/send_null/send_null.py
lyriccoder: If you really need... In future if you add smth else, you can make a mistake. It is more readable

aibolit/patterns/send_null/send_null.py
lyriccoder: no, set are unordered:

` a_set = {6,7,8,9}`
`{8, 9, 6, 7}`

test/patterns/send_null/test_send_null.py
lyriccoder: we found new examples

aibolit/patterns/send_null/send_null.py
lyriccoder: in this case we will have side effect. It is an anti-pattern

aibolit/patterns/send_null/send_null.py
aravij: Ok then

aibolit/patterns/send_null/send_null.py
aravij: I mean if we will use `list` for storing line number, they will probably come up all in right order. No need for `set` or `sorted`.

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,37 @@ class SendNull:
...
-                if not isinstance(i, list):
-                    if hasattr(i, '_position'):
-                        if i.position:
-                            return i._position.line
-                        else:
-                            for j in i.children:
-                                position = self.__find_position_recursively(j)
-                                if position:
-                                    return position
+    def __is_null(self, val):
```
aravij, line 12: I hardly can find this function neither readable nor maintainable. Hope you too.

aibolit/patterns/send_null/send_null.py
```diff
@@ -9,59 +9,37 @@ class SendNull:
...
+        has_ternary_true_value = hasattr(val, 'value')
+        if has_ternary_true_value:
+            is_ternary_true_str = isinstance(val.value, str)
+            if is_ternary_true_str:
+                if val.value == 'null':
+                    return True
```
acheshkov, line 18:
```suggestion
    def __is_null(self, val):
        if not hasattr(val, 'value'): 
            return False
        if not  isinstance(val.value, str): 
            return False
        if val.value != 'null': 
            return False
        return True
```

aibolit/patterns/send_null/send_null.py
acheshkov: @lyriccoder, @aravij ask you the example

aibolit/patterns/send_null/send_null.py
acheshkov: @lyriccoder I think you can write it more elegant

aibolit/patterns/send_null/send_null.py
aravij: @lyriccoder  If there exist such example, that you cannot simply check for node is `LITERAL` or its `value` being something else than string, than we better add it to the tests. After all I still consider that checking for node type `LITERAL` is enough.

aibolit/patterns/send_null/send_null.py
lyriccoder: this line will fail: 
`if not  isinstance(val.value, str): ` if val doesn't have value field. Your code will fail the program

aibolit/patterns/send_null/send_null.py
acheshkov, line 12: please, add typings

aibolit/patterns/send_null/send_null.py
acheshkov: `if val doesn't have value field` I will exit with False before

test/patterns/send_null/test_send_null.py
```diff
@@ -26,7 +26,7 @@ def test_simple_invocation(self):
 
     def test_more_method_invocations(self):
         lines = self.method_send_null_finder.value(Path(self.dir_path, 'SequenceFile.java'))
-        self.assertEqual(lines, [1186, 1201, 1217, 3285, 3298, 3367, 3537, 3550])
+        self.assertEqual(lines, [1097, 1186, 1201, 1217, 3285, 3298, 3367, 3537, 3550])
```
acheshkov, line 29: why do we change asset expression?

lyriccoder: new cases were found since I was not looking for Literal only

aibolit/patterns/send_null/send_null.py
lyriccoder: fixed

acheshkov:
@rultor merge

PR #577, opened 2020-07-28
Redundant_catch refactored 

Vitaly-Protasov:
Solved in case of #528 

Also test was change because of previous error (you can check the sourse code and make sure that this code structure should be detected too

### DIFF
--- a/test/patterns/redundant_catch/test_redundant_catch.py
+++ b/test/patterns/redundant_catch/test_redundant_catch.py
@@ -104,4 +104,4 @@ class RedundantCatchTest(TestCase):
         """
         pattern = RedundantCatch()
         lines = pattern.value(os.path.dirname(os.path.realpath(__file__)) + '/Cache.java')
-        self.assertEqual(lines, [])
+        self.assertEqual(lines, [393])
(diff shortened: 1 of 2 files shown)


aravij:
@acheshkov @lyriccoder  I have finished the task, so please take additional look and merge if it is ok.

test/patterns/redundant_catch/test_redundant_catch.py
```diff
@@ -104,4 +104,4 @@ def test_fake_try_in_lambda(self):
         """
         pattern = RedundantCatch()
         lines = pattern.value(os.path.dirname(os.path.realpath(__file__)) + '/Cache.java')
-        self.assertEqual(lines, [])
+        self.assertEqual(lines, [393])
```
lyriccoder, line 107: no, it's wrong behaviour. W have lambda. There is no another way to handle exception in lambda. There was a test to check that. If we have try in lambda, we should ignore it

Vitaly-Protasov: @lyriccoder  I FIXED IT!

acheshkov:
@rultor merge

aravij:
@rultor why not merging?

PR #578, opened 2020-07-29
Refactor nested blocks.

aravij:
Delete nested loops and use nested blocks pattern drectly.


aravij:
There is one test failing. Are we considering a pattern with nested if like this:
```java
if(x == 0) {
    doSmth();
    if(y == 0) {
        doSmth();
    }
    doSmth();
}
```
It does seems to me a direct nesting like in all other test examples with cycles.

acheshkov:
@rultor merge

acheshkov:
@aravij please take a look

aravij:
@acheshkov The merge will fail anyway, because unit tests are failing. I posted above a comment about current pattern. How does this pattern should behave? I can't extract this information from comments. There is only one such example. is it correct or not?

acheshkov:
> @acheshkov The merge will fail anyway, because unit tests are failing. I posted above a comment about current pattern. How does this pattern should behave? I can't extract this information from comments. There is only one such example. is it correct or not?

@aravij Sorry, did not realize that it's a question.

acheshkov:
> There is one test failing. Are we considering a pattern with nested if like this:
> 
> ```java
> if(x == 0) {
>     doSmth();
>     if(y == 0) {
>         doSmth();
>     }
>     doSmth();
> }
> ```
> 
> It does seems to me a direct nesting like in all other test examples with cycles.

Yes, let's consider this as a pattern match

aravij:
@acheshkov tests are fixed. Please approve and merge.

acheshkov:
@rultor merge

PR #590, opened 2020-08-04
Rewrite class decomposition.

aravij:


test/utils/test_ast_node.py
```diff
@@ -43,7 +43,8 @@ def test_fields(self):
         self.assertEqual(java_class.modifiers, set())
         self.assertEqual(java_class.documentation, "/**\n* Some documentation\n*/")
 
-        fields_names = {field.name for field in java_class.fields}
+        # consider each field declaration declares single field
+        fields_names = {field.names[0] for field in java_class.fields}
```
lyriccoder, line 47: is it possible to add test when we cannot decompose class:
1) it has LCOM1, so it should return the class itself
2) if nothing to decompose (empty class or interface)

acheshkov:
@rultor merge

PR #595, opened 2020-08-05
Add list() in model's output

angusev:
@lyriccoder please check if this commit raises no conflicts


lyriccoder:
why do u need such a change?

PR #604, opened 2020-08-07
Aibolit WP: split long sentence. 

### LINKED ISSUE #603: Aibolit WP: write it better
https://github.com/cqfn/aibolit/blob/7ba9e693d551f110c92521af02b5fa09f33163bc/wp/sections/usage_scenarios.tex#L3

write this sentence better

acheshkov:
closes #603 
the long sentence is split and written simpler.

### DIFF
--- a/wp/sections/usage_scenarios.tex
+++ b/wp/sections/usage_scenarios.tex
@@ -1,7 +1,7 @@
 \subsection{Aibolit Index}
 
-In addition to the recommendation functionality, we designed an \textit{"Aibolit Index"}. 
-The Aibolit Index is useful to metric the overall quality of some project.
+In addition to the recommendation functionality, we proppose \textit{Aibolit Index}, 
+a score that measures the overall quality of a given software project.
 It is is a single number, with following properties: 
 \begin{itemize}
 \item[(i)] the more patterns are suggested to fix, the higher Aibolit Index;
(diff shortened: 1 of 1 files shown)


wp/sections/usage_scenarios.tex
```diff
@@ -1,6 +1,7 @@
 \subsection{Aibolit Index}
 
-In addition to the recommendation functionality, we designed a feature to measure the overall quality of developer's source code, \textit{Aibolit Index}.
+In addition to the recommendation functionality, we designed an \textit{"Aibolit Index"}.
```
KatGarmash, line 3: I actually don't think you should split this sentence.

I'd just rewrite a bit: 
In addition to the recommendation functionality, we prppose \textit{Aibolit Index}, a score that measures the overall quality of developer's source code.

acheshkov: I'm okay even with the previous version of the sentence, except this part `of developer's source code`. I suggest this refactoring, as I understand the professor's comments.

KatGarmash: Ah then maybe:

"""
In addition to the recommendation functionality, we proppose \textit{Aibolit Index}, a score that measures the overall quality of a given software project.
"""

I think "a given" sounds better than "some" in this context.

acheshkov: I like it

PR #607, opened 2020-08-08
Scope graph

### LINKED ISSUE #570: Copmuted field 'statements_by_scope'
To provide seamless functionality with `JavalangImproved` we need to be able to provide all method statements grouped by scope.

aravij:
Fix #570 
Scope graph provides tree like structure with code scopes as nodes. Each scope has a list of its statements. Handles lambdas inside expressions, if and cycle conditions, etc. Anonymous class are not supported currently.

While it is said 15 files were changed, half of it was just moved in another directory. It is done to make separate directory for `ast_framework` tests. Most important changes are in `scope.py` and `scope_extractors.py` files. This new functionality tested with `test_scope.py` and `ScopeTest.java` files, which is quite large, just to cover a lot of cases.


acheshkov:
@rultor merge

Issue #468, opened 2020-06-30
Change API for work with class components


aravij:
Now patterns output error line numbers for a file, but we need to make them output error line numbers for all class components. Code, handling patterns need to be properly changed.

0crat:
@acheshkov[/z](https://www.0crat.com/u/acheshkov) please, pay attention to this issue

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 4@8366e303-9dc5-4679-a29a-c61b61f06cfb, thread: PQ-CT5UB7XAP
com.jcabi.xml.StrictXML[124] java.lang.IllegalArgumentException: 2 error(s) in XML document: -1:-1: cvc-pattern-valid: Value 'gh:Evaluate pattern on arbitary AST' is not facet-valid with respect to pattern '[a-z]{2}:[A-Z0-9a-z.\-#/]+' for type 'milestoneID'.;-1:-1: cvc-complex-type.2.2: Element 'successor' must have no element [children], and the value must be valid.
```

[0.53.15](https://github.com/zerocracy/farm): CID: [d83529bb-7670-42fd-89d4-9df27e9757af](https://www.0crat.com/footprint/CT5UB7XAP/d83529bb-7670-42fd-89d4-9df27e9757af), Type: "Job milestoned"

PR #611, opened 2020-08-12
[#468] New dataset on LCOM components

aravij:
- For each file in dataset we extract all classes and for each class extract all LCOM components. Information about class and component is stored in dataset in fields `class_name` and `component_index`.
 - Almost all patterns are now accept `AST` instead of file path. Exception:
   -  `var_diff` (P20)  - not refactored
   - "P28", "P9" - text based patterns
 - Only NCSS metric present in new dataset. Cognitive complexity needs refactoring.
 - Dataset calculation procedure changed. It now has a timout in 60 seconds for single file processing (can be changed with `-t` CLI parameter). All exceptions are printed right in the console. Failure of file processing does not stop the rest of computation. Progress bar is added.

Most of the files you will find here are patterns and their tests.


.gitignore
```diff
@@ -29,3 +29,9 @@ sphinx
 *.ist
 wp.pdf
 scripts/_tmp
+catboost_info/catboost_training.json
+catboost_info/learn_error.tsv
+catboost_info/test_error.tsv
```
lyriccoder, line 34: isn't it better to ignore all catboost folder? Moreover ,it should be deleted with the previous implementation

test/patterns/supermethod/test_supermethod.py
```diff
@@ -20,42 +20,64 @@
...
+        filepath = self.current_directory / "NestedClass.java"
+        ast = AST.build_from_javalang(build_ast(filepath))
+        pattern = SuperMethod()
+        lines = pattern.value(ast)
+        self.assertEqual(len(lines), 1)
 
     def test_constructor(self):
-        file = str(Path(self.cur_file_dir, 'Constructor.java'))
-        self.assertEqual(len(self.testClass.value(file)), 3)
+        filepath = self.current_directory / "Constructor.java"
```
lyriccoder, line 72: why don't u like `Path(self.cur_file_dir, 'Constructor.java')`?
I see u are using 2 interfaces: `Constructor` and `operator /`. Shouldn't we use only one?

.gitignore
aravij: Well it don't get deleted after I launch integration tests.

test/patterns/supermethod/test_supermethod.py
aravij: I wouldn't call two methods of a class two different interfaces. I create a `Path` object with constructor and then use its methods and operator to get other paths. Anyway it looks more simpler to me.

scripts/04-find-patterns.py
```diff
@@ -20,208 +20,157 @@
...
+            f"while process only allowed to occupy {allowed_cores_qty} cores."
+            "You may have a slowdown due to large number of processes.",
+            file=stderr,
+        )
+
+    if args.timeout == 0:
+        args.timeout = None
+
+    default_dataset_directory = Path(".", "target", "04").absolute()
+    dataset_directory = getenv("TARGET_FOLDER", default=default_dataset_directory)
```
lyriccoder, line 155: make 
```
if target_folder:
    os.chdir(target_folder)
```, 

otherwise target folder won't work

lyriccoder (changes requested):
add exceptions handler and grouping them by error and pattern name
make os.chdir in 04-patterns.py

scripts/04-find-patterns.py
aravij: Changing directory might be quite dangerous. Anyway I use this directory in the following way `Path(dataset_directory, "04-find-patterns.csv")`, so dataset is created in right directory.

Issue #623, opened 2020-08-20
sum of NCSS of all components must be equal to the ncss of the whole file


lyriccoder:
sum of NCSS of all components must be equal to the ncss of the whole file.
 Now it differs significantly for all files.

Example is here:
[Configuration.java.txt](https://github.com/cqfn/aibolit/files/5103402/Configuration.java.txt)

1211 (all components) vs 1502 (the whole file)

PR #624, opened 2020-08-20
Test for ncss value for all components

lyriccoder:
Please specify the following in the description:

Test for https://github.com/cqfn/aibolit/issues/623

It should be fixed in https://github.com/cqfn/aibolit/pull/611


aibolit/config.py
```diff
@@ -174,6 +174,8 @@ def get_patterns_config():
             },
             "patterns_exclude": [
                 "P27",  # empty implementation
+                "P20_5", "P20_7", "P20_11",  # wasn't refactored yet
+                "P28", "P9",  # patterns based on text cannot accept arbitrary AST
             ],
-            "metrics_exclude": []
+            "metrics_exclude": ["M1", "M3_1", "M3_2", "M3_3", "M3_4", "M5", "M6", "M7"]
```
acheshkov, line 180: why we exclude metrics?

aravij: They have been excluded in my PR. I did it, because they are not refactored. So we only have NCSS and Cognitive.

aravij:
There was some failures with NCSS test. It was due to presence of CLASS_DECLARATION node in all components and presence of nested classes, which was not particapated in decomposition.

To bypass nested classes I calculate NCSS of only class methods and fields instead of calculating whole class NCSS. To bypass CLASS_DECLARATION problem I add number of components to the sum of methods and fields NCSS. It compensate presence of this noe in components.

lyriccoder:
@acheshkov  could you please review? I can't review/accept my own PR

acheshkov (approved):

acheshkov:
@rultor merge

PR #629, opened 2020-08-25
Add ability to not decompose classes on env var present.

### LINKED ISSUE #628: Add ability to create dataset without decomposition.
This can be done with help of environment variable `LCOM_DECOMPOSITION`. If it set to "True" or "1" decomposition is applied. If it is set to "False" or "0" it is not applied. By default decomposition is applied.

aravij:
Fix #628

If there is environment variable `LCOM_DECOMPOSITION` and it is equal to "False" or "0" all classes processed as is without LCOM decomposition. If it is equal to "True", "1" or not set at all decomposition is applied. In all other cases decomposition is still applied, but warning message is printed.


lyriccoder (approved):
ok

acheshkov:
@rultor merge

aravij:
@acheshkov those bots 0pdd and zerocrat just post their stack traces and nothing else. We better delete them from our repository. They don't help, just distract.

acheshkov:
> @acheshkov those bots 0pdd and zerocrat just post their stack traces and nothing else. We better delete them from our repository. They don't help, just distract.

@aravij let's discuss it on a meeting

PR #633, opened 2020-08-26
Multitargeting

### LINKED ISSUE #632: Mutitargeting.
Add ability to train against different metrics:
 - RFC
 - Fanout
 - Max diameter

aravij:
Fix #632 

Added ability to choose target metric to be trained against.
Currently supported:
 - Cognitive complexity
 - Fan out
 - Max methods diameter
 - Response for a class

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -120,10 +120,24 @@ def train():
         required=False,
         default=False
     )
+    parser.add_argument(
+        "--target-metric",
+        choices=["rfc", "fanout", "diameter", "cognitive"],
+        default="cognitive",
+        help="Select target metric. Available metrics: rfc, fanout, diameter, cognitive. "
+             "Default is cognitive."
+    )
     args = parser.parse_args(sys.argv[2:])
     if not args.skip_collect_dataset:
         collect_dataset(args)
-    train_process()
+
+    metric_name_to_code = {
+        "rfc": "M9",
+        "fanout": "M10",
+        "diameter": "M6",
+        "cognitive": "M4",
+    }
+    train_process(metric_name_to_code[args.target_metric])
 
 
 def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_metric=False):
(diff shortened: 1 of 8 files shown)


aibolit/__main__.py
```diff
@@ -120,10 +120,24 @@ def train():
...
     if not args.skip_collect_dataset:
         collect_dataset(args)
-    train_process()
+
+    metric_name_to_code = {
+        "rfc": "M9",
+        "fanout": "M10",
+        "diameter": "M6",
+        "cognitive": "M4",
+    }
```
acheshkov, line 139: If possible, I'd add this data to aibolit/config.py

aibolit/model/model.py
```diff
@@ -54,7 +54,7 @@ def scale_dataset(
     metrics_codes_set = [x['code'] for x in config['metrics']]
     exclude_features = set(config['patterns_exclude']).union(set(config['metrics_exclude']))
     used_codes = set(features_conf['features_order'])
-    used_codes.add('M4')
+    used_codes |= {'M4', 'M6', 'M9', 'M10'}
     not_scaled_codes = set(patterns_codes_set).union(set(metrics_codes_set)).difference(used_codes).difference(
```
lyriccoder, line 58: can we just add a parameter, 1 parameter target. And then we will add
` used_codes.add(target)`

lyriccoder (changes requested):
Set target as a parameter to scale function

aibolit/model/model.py
aravij: Done.

lyriccoder (approved):
ok

aibolit/__main__.py
aravij: It is possible, but this dict is specific to command line arguments of train, so I would keep it close to that method. Anyway it use different names than in config.

lyriccoder:
@acheshkov merge it pls

acheshkov:
@rultor merge

PR #635, opened 2020-09-15
Extract semantic

### LINKED ISSUE #634: Extract semantic.
Extracts semantic from methods.

aravij:
Fix #634

Extracts semantic from methods. First step for Baseline. Currently only plain statements are supported. All statements, which alow nesting will be added in following PR.

### DIFF
--- a/test/extract_method_baseline/SimpleMethods.java
+++ b/test/extract_method_baseline/SimpleMethods.java
@@ -1,0 +1,38 @@
+class SimpleMethods {
+    void assertStatement(int x) {
+        assert x == 0;
+    }
+
+    int returnStatement(int x) {
+        return x;
+    }
+
+    void expression(int x) {
+        x++;
+    }
+
+    void throwStatement(Error x) {
+        throw x;
+    }
+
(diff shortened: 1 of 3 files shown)


test/extract_method_baseline/SimpleMethods.java
```diff
@@ -0,0 +1,38 @@
+class SimpleMethods {
```
acheshkov, line 1: why we have so a simple test? could we use more sophisticated methods? Maybe we can take it from the paper?

aravij: Then it will be more close to integration tests. We better keep unit tests simple for easy debug. If you want some integration tests we need to understand what exactly they should solve? Is it just fixing behavior on some production code? Or something more?

aravij: I will add more complex tests later when complex statements will be supported.

Vitaly-Protasov (approved):

acheshkov (approved):

acheshkov:
@rultor merge

PR #676, opened 2025-05-14
Updated flatten() in __main__.py

Vitalo-bit:
I've added input type annotation List[List[Any]] to specify that the function takes a list of lists

Also I've added return type annotation List[Any] to specify that the function returns a flattened list

Even if it is a small change it Improves code readability and makes the code more maintainable

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -140,7 +140,7 @@ def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_
         val = value_dict['make']().value(ast)
         if not is_metric:
             input_params[acronym] = len(val)
-            code_lines_dict['lines_' + acronym] = val
+            code_lines_dict["lines_" + acronym] = val
         else:
             input_params[acronym] = val
     except Exception:
@@ -151,7 +151,7 @@ def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_
         )
 
 
-def flatten(l):
+def flatten(l: List[List[Any]]) -> List[Any]:
     return [item for sublist in l for item in sublist]
 
 


yegor256:
@Vitalo-bit thanks for the contribution, but we need a test to validate this change. Without a test, your changes may be rolled back in the future.

yegor256:
@Vitalo-bit please, resolve merge conflicts

ivanovmg:
> @Vitalo-bit thanks for the contribution, but we need a test to validate this change. Without a test, your changes may be rolled back in the future.

In Python there is no opportunity to test it using unit tests. The only option is to use `mypy --strict` to force every single function to have type annotation. However, for this project this would be a long journey, definitely beyond the scope of this PR.

aibolit/__main__.py
```diff
@@ -140,7 +140,7 @@ def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_
         val = value_dict['make']().value(ast)
         if not is_metric:
             input_params[acronym] = len(val)
-            code_lines_dict['lines_' + acronym] = val
+            code_lines_dict["lines_" + acronym] = val
```
ivanovmg, line 143: I suppose, this change is out of this PR scope.

ivanovmg:
CI is failing, maybe due to use of old base commit. I suggest that you rebase on top of master.

aibolit/__main__.py
```diff
@@ -151,7 +151,7 @@ def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_
         )
 
 
-def flatten(l):
+def flatten(l: List[List[Any]]) -> List[Any]:
```
ivanovmg, line 154: Instead of Any consider using generics.

ivanovmg, line 154: Something like this:
```suggestion
def flatten(l: List[List[T]]) -> List[T]:
```

But also need to declare `T`:
``` python
from typing import TypeVar
T = TypeVar('T')
```

ivanovmg:
@Vitalo-bit can you please rebase it on top of master and force push?

PR #686, opened 2025-05-18
resolve #685 add `.xcop` to skip virtual environment analyze by `xcop`

### LINKED ISSUE #685: `make` fails due to `xcop`
When running the project with a Python 3.11 or 3.7 virtual environment located inside the project directory, the make command fails because `xcop` attempts to read an unintended XML file from the virtual environment:
`No such file or directory @ rb_sysopen - /home/lbc/Documents/aibolit/.venv/lib/python3.11/site-packages/setuptools/command/launcher`

This occurs during execution of: `xcop $$(find . -name '*.xml')`
In the following path inside the virtual environment: `/home/lbc/Documents/aibolit/.venv/lib/python3.11/site-packages/setuptools/command/` exists a file named: `'launcher manifest.xml'`

Due to the space in the filename, `xcop` incorrectly splits it into two arguments (launcher and manifest.xml) and fails.
In my opinion, files from the virtual environment should not be processed at all, since they are unrelated to the project’s source code and are generated automatically by package managers.

Moving the virtual environment outside the project directory avoids the issue entirely. In this case, make runs successfully

literally-bug-creator:


literally-bug-creator:
Resolves #685

literally-bug-creator:
@yegor256 Can you please do a review of it?

README.md
```diff
@@ -327,8 +327,14 @@ Install the following packages if you don't have:
 apt-get install ruby-dev libz-dev libxml2
 ```
 
-Then, you fork the repo and make the changes. Then, you make
-sure the build is still clean, by running:
+This project does not include a virtual environment by default.
+If you're using one (e.g., .venv, venv), update the .xcop file to exclude it:
```
yegor256, line 331: @literally-bug-creator what's the point in this notice, if you created `.xcop` file? It will automatically work for everybody.

literally-bug-creator: > @literally-bug-creator what's the point in this notice, if you created `.xcop` file? It will automatically work for everybody.

You're right, `.xcop` should handle it. I just added the notice because Python devs often name their virtual environments differently. If someone uses a name that's not excluded in `.xcop`, it could cause issues unless they update the config. The notice is just to help avoid that.

PR #699, opened 2025-05-21
Resolve #697: Temporary fix: limit `xcop`

### LINKED ISSUE #697: `xcop` ignores `.xcop` configuration and raises error when using virtual environment
While working on the project, I encountered an issue where `xcop` fails with a file not found error when executed from within a Python virtual environment, despite the `.xcop` configuration file specifying that the `.venv/` directory should be excluded. Ref #685 

# Notes:
1. The `.xcop` file is located at the project root.
2. However, when using Python 3.11, I consistently get the following error **after installing `setuptools` into the virtual environment**

# Reproduction steps:
1. Create a virtual environment using Python 3.11:
```python3.11 -m venv .venv```
2. Run `make xcop`

# Actual result:
`No such file or directory @ rb_sysopen - /home/lbc/Documents/aibolit/.venv/lib/python3.11/site-packages/setuptools/command/launcher`

# Expected result:

**`xcop` should ignore the `.venv` directory as specified in the `.xcop` configuration:**
```xcop
--exclude=.venv/**
```

**and should not attempt to scan files inside it.**

literally-bug-creator:
This PR (resolve #697) updates the `Makefile` to run `xcop` only on the `aibolit/` and `scripts/` directories.
This helps avoid issues with virtual environments being unintentionally scanned.

**_Note: the `.xcop` ignore config does not seem to work as expected — it doesn't ignore the virtual environment directory. This needs further investigation._**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,8 @@ it:
 	./test/integration/test_recommend.sh
 
 xcop:
-	xcop $$(find . -name '*.xml')
+	xcop $$(find ./aibolit -name '*.xml')
+	xcop $$(find ./scripts -name '*.xml')
 
 flake8:
 	python3 -m flake8 aibolit test scripts setup.py --exclude scripts/target/*


literally-bug-creator:
> @literally-bug-creator I added `.xcop` config file, see [63934c4](https://github.com/cqfn/aibolit/commit/63934c41287bd5693098a9c3a5a69d3be3c58dd0) Does it help?

@yegor256 Unfortunately, no. I think it's worth using PR until this problem is solved, because not being able to use the virtual environment inside the project (during build time), stops work on other tasks

PR #711, opened 2025-05-23
Fix ruff workflow

### LINKED ISSUE #710: Fix ruff/black workflow in CI
It is necessary to fix `ruff/black` workflow in CI.

              @ivanovmg if you can fix "black" and "ty" workflows, would be excellent. This is most important now.

_Originally posted by @yegor256 in https://github.com/cqfn/aibolit/issues/708#issuecomment-2903539236_

ivanovmg:
This PR attempts to fix `ruff` CI workflow.

For convenience I added `make ruff` step, which is useful during development.
The workflow complained about one file, which I refactored to fix the problem.
- https://github.com/cqfn/aibolit/blob/master/aibolit/metrics/loc/loc.py

Closes #710

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,9 @@ flake8:
 pylint:
 	python3 -m pylint aibolit test scripts --ignore=scripts/target
 
+ruff:
+	python3 -m ruff check .
+
 sphinx:
 	rm -rf sphinx html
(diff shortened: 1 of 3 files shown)


yegor256:
@ivanovmg thanks!

PR #716, opened 2025-05-23
Use venv for ty workflow

### LINKED ISSUE #715: `ty` workflow needs to run from venv
`ty` workflow needs to run from a virtual environment as described here: https://github.com/astral-sh/ty/blob/main/docs/README.md#python-environment

However, at the moment is not run withing a venv. Thus, we see multiple errors related to not being able to import third party packages.

Xref: #712

ivanovmg:
Use venv for ty workflow:
https://github.com/astral-sh/ty/blob/main/docs/README.md#python-environment

In this PR we activate venv in each relevant step since `github` actions run each script from a different shell.
Closes #715


ivanovmg:
@yegor256 it does fix the issue.
Compare with the previous `ty` run: https://github.com/cqfn/aibolit/actions/runs/15207609077/job/42774241498 (101 errors).
Current version has 53 typing-related errors and none of them is related to inability to import a third party library.

yegor256:
@ivanovmg thanks!

PR #723, opened 2025-05-25
FIX: handle non-OK status code when requesting aibolit version

### LINKED ISSUE #722: Handle potential error when requesting last aibolit version
Today `make` pipeline failed due to this error:

```
└➜ python aibolit/__main__.py
> /home/maxim/github/aibolit/aibolit/__main__.py(891)get_versions()
-> releases = json.loads(requests.get(url, timeout=15).content)['releases']
(Pdb) resp = requests.get(url)
(Pdb) resp
<Response [406]>
(Pdb) resp.text
'python-requests/2.32.3 User-Agents are currently blocked from accessing JSON release resources. A cluster is apparently crawling all project/release resources resulting in excess cache misses. Please contact admin@pypi.org if you have information regarding what this software may be.'
(Pdb)
```

I suppose that the section related to versions comparison should be reconsidered.

Consider raising error when status code is not OK.

_Originally posted by @ivanovmg in https://github.com/cqfn/aibolit/issues/721#issuecomment-2907686753_

ivanovmg:
This change handles non-OK status code when requesting the latest aibolit version from PyPI.

`response.raise_for_status()` succeeds if status code is less than 400, otherwise raises `HTTPError`, which we then handle on the caller's side https://docs.python-requests.org/en/latest/_modules/requests/models/#Response.raise_for_status

Closes #722

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -887,7 +887,9 @@ def run_thread(files, args):
 
 def get_versions(pkg_name):
     url = f'https://pypi.python.org/pypi/{pkg_name}/json'
-    releases = json.loads(requests.get(url, timeout=15).content)['releases']
+    response = requests.get(url, timeout=15)
+    response.raise_for_status()
+    releases = json.loads(response.content)['releases']
     return sorted(releases, key=parse_version, reverse=True)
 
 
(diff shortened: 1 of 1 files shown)


yegor256:
@ivanovmg thanks!

PR #727, opened 2025-05-26
Add codecov.yml workflow

ivanovmg:
This PR adds `codecov.yml`, which is dedicated to publishing code coverage to codecov.io

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
+# SPDX-License-Identifier: MIT
+---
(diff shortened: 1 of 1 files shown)


ivanovmg:
@yegor256 can you please clarify whether the `codecov` workflow must trigger on merge/push to master branch, or in pull request as well?

Currently I cannot test it as I set the trigger on merge to master, like in your example: https://github.com/objectionary/lints/blob/master/.github/workflows/codecov.yml
For testing purposes we can alternatively add trigger on pull request, check that it works and once it does, rollback to trigger on push to master only.

What do you think?

yegor256:
@ivanovmg you did it right, thanks! now, let's see how it works...

Issue #736, opened 2025-05-27
`ClassicGetter` pattern does not catch getters with this reference


ivanovmg:
`ClassicGetter` pattern treats this piece of code as OK:
```
class Dummy {
    private int value;
    public int getValue() {
        return this.value;
    }
}
```
However, this is still a classic getter.

0pdd:
@ivanovmg the puzzle [#762](https://github.com/cqfn/aibolit/issues/762) is still not solved; solved: [#758](https://github.com/cqfn/aibolit/issues/758).

0pdd:
@ivanovmg all 2 puzzles are solved here: [#758](https://github.com/cqfn/aibolit/issues/758), [#762](https://github.com/cqfn/aibolit/issues/762).

ivanovmg:
Closing as all puzzles were resolved.

PR #737, opened 2025-05-27
Add a failing test on ClassicGetter

ivanovmg:
This PR adds a failing test on classic getter pattern where `this` reference is used.

Xref: #736


ivanovmg:
@yegor256 here I added a failing test - it does indeed fail. Will it be fine to merge it and then fix implementation separately?

yegor256:
@ivanovmg you can use PDD, leaving the marker/puzzle in the code. See: https://www.yegor256.com/2010/03/04/pdd.html

yegor256:
@ivanovmg thanks! here it is: https://github.com/cqfn/aibolit/issues/758

Issue #776, opened 2025-05-29
test_bidirect_index.py puzzle resolution is required


0pdd:
The puzzle `772-42868a9b` from #772 has to be resolved: 

https://github.com/cqfn/aibolit/blob/925185b87eca5120f0f7bf33df680bad9ce799af/test/patterns/bidirect_index/test_bidirect_index.py#L58-L60

The puzzle was created by Yegor Bugayenko on 29-May-25. 

Estimate: 15 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

yegor256:
@ivanovmg well, the `bidirectIndexHiddenScopeFalse()` method indeed matches the pattern, doesn't it?

ivanovmg:
@yegor256 as I look at this example, it becomes less obvious to me :)
Technically, the provided example https://github.com/cqfn/aibolit/blob/master/test/patterns/bidirect_index/BidirectIndexHiddenScope.java does not compile due to variable `i` not defined.

Take a look at a curated example with while loop, where we declare the variable in the class:
```
class WithWhileLoop {
    private int i;
    void method() {
        i = 0;
        while(true) {
            int i = 0;
            ++i;
        }
        --i;
    }
}
```

It does not compile either since `--i` is not reachable:
```
ERROR!
/tmp/ZggDULNwC5/Main.java:9: error: unreachable statement
        --i;
        ^
1 error
```

yegor256:
@yegor256 aibolit doesn't care whether a Java file is compilable or not. It only cares about the presence of a pattern.

It was my mistake above. Both methods _don't_ have the pattern.

ivanovmg:
@yegor256 can you mark it as a bug? To avoid me having to deal with the mean zerocrat? :)

yegor256:
@ivanovmg 0pdd should automatically add "bug" labels. Let me check why it's not happening.

Issue #684, opened 2025-05-16
Limitations in test maintainability and debugging with current `unittest` framework


literally-bug-creator:
`pytest` is fully compatible with existing `unittest-style` tests, which means we can gradually migrate test files without rewriting all tests at once.   

Benefits of switching to `pytest` include: 
- More readable and concise test syntax  
- Built-in support for parametrized testing  
- Rich plugin ecosystem (e.g. pytest-xdist, pytest-cov, etc.)  
- Easier setup and teardown via fixtures  
- Better error reporting and debugging experience

     

This could improve developer experience, make test code more maintainable, and unlock more advanced testing features.

yegor256:
@literally-bug-creator sure, good idea

PR #800, opened 2025-06-02
Implement `IncompleteFor` pattern. Closes #140

### LINKED ISSUE #140: incomplete_for pattern
Here it is:

```java
class Foo {
  void bar() {
    for(int x = 0; i < 10;) { // here!
      x++;
    }
  }
}
```

A `for` loop must have all three parts defined. Very often programmers skip one, two or all three of them. This definitely is a pattern.

AntonProkopyev:
Closes #140

### DIFF
--- a/test/patterns/incomplete_for/test_incomplete_for.py
+++ b/test/patterns/incomplete_for/test_incomplete_for.py
@@ -47,3 +47,21 @@ class IncompleteForTestCase(TestCase):
             IncompleteFor().value(ast),
             "Should match empty for",
         )
+
+    def test_incomplete_for_nested_for(self):
+        filepath = self.dir_path / "NestedFor.java"
+        ast = AST.build_from_javalang(build_ast(filepath))
+        self.assertEqual(
+            [6, 7],
+            IncompleteFor().value(ast),
+            "Should match nested for loops",
+        )
+
+    def test_incomplete_for_complete_for(self):
+        filepath = self.dir_path / "CompleteFor.java"
+        ast = AST.build_from_javalang(build_ast(filepath))
+        self.assertEqual(
+            [],
+            IncompleteFor().value(ast),
+            "Should not match complete for loop",
+        )
--- a/test/patterns/incomplete_for/CompleteFor.java
+++ b/test/patterns/incomplete_for/CompleteFor.java
@@ -1,0 +1,9 @@
+// SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
+// SPDX-License-Identifier: MIT
+
+class IncompleteForCompleteFor {
+  void method() {
+    for(int i = 0; i < 10;++i) {
+    }
+  }
+}
--- a/test/patterns/incomplete_for/NestedFor.java
+++ b/test/patterns/incomplete_for/NestedFor.java
@@ -1,0 +1,13 @@
+// SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
+// SPDX-License-Identifier: MIT
+
+class IncompleteForNestedFor {
+  void method() {
+    for(int i = 0; i < 10;) {
+      for(int j = 0; j < 10;) {
+        ++j;
+        ++i;
+      }
+    }
+  }
+}


test/patterns/incomplete_for/test_incomplete_for.py
```diff
@@ -0,0 +1,49 @@
...
+from aibolit.ast_framework.ast import AST
+from aibolit.patterns.incomplete_for.incomplete_for import IncompleteFor
+from aibolit.utils.ast_builder import build_ast
+
+
+class IncompleteForTestCase(TestCase):
+    dir_path = Path(os.path.realpath(__file__)).parent
+
+    def test_incomplete_for_no_init_part(self):
+        filepath = self.dir_path / "NoInitPart.java"
```
ivanovmg, line 16: @AntonProkopyev it may be a good opportunity to develop tests, which do not touch a file system. Instead we can define string values of the file content and instantiate `ast` object from the string content. This way the tests will be faster, but also we will have test fixture and the actual test in the same place. Take a look at this example https://github.com/cqfn/aibolit/pull/737/files#diff-a61e0d942f09a1332a1a4645544e484334c1e0dcd32e34b1b3be0504b60c51db

@yegor256 what are your thoughts in this regard?

ivanovmg: @AntonProkopyev may I suggest that we add one more case where two or more lines match the pattern? I may think of a for loop inside a for loop.

ivanovmg (commented):
@AntonProkopyev thank you for the PR! I have a couple of comments. Please take a look.

ivanovmg:
@AntonProkopyev if you use "Closes #XXX" in the PR description, then it will tell github that once this PR is merged, the issue must get closed automatically.

aibolit/patterns/incomplete_for/incomplete_for.py
```diff
@@ -0,0 +1,23 @@
...
+    Finds all for statements, which don't have at least one control part
+    """
+
+    def value(self, ast: AST) -> List[LineNumber]:
+        lines: set[int] = set()
+        for for_control in ast.get_proxy_nodes(ASTNodeType.FOR_CONTROL):
+            if not all((for_control.init, for_control.update, for_control.condition)):
+                parent = for_control.parent
+                if parent is None:
+                    raise RuntimeError("For control without For statement parent")
```
ivanovmg, line 21: @AntonProkopyev can you please also add a test to cover this line? It is not clear to me right now at which conditions this exception is thrown.

test/patterns/incomplete_for/test_incomplete_for.py
AntonProkopyev: @ivanovmg Thank you for your observation. I have some doubts about embedding Java classes into Python test code. As long as they remain separate files, we can try to compile them separately, which would help verify that the file in the test is syntactically correct. A command may look like:
```bash
find test/ -name "*.java" -print -exec javac {} \;
```

aibolit/patterns/incomplete_for/incomplete_for.py
AntonProkopyev: @ivanovmg `make mypy` complains that the `parent` attribute may return `None` instead of an `ASTNode`. This situation should probably never happen, but if I add an `assert`, Ruff will complain with [S101](https://docs.astral.sh/ruff/rules/assert/), because an `assert` will not work in some cases.

AntonProkopyev:
@ivanovmg Thank you for taking the time to review my PR.

aibolit/patterns/incomplete_for/incomplete_for.py
ivanovmg: @AntonProkopyev ok, thank you for letting me know. From my perspective, this is a legitimate case for silencing the checker via `# noqa: S101`, especially considering that `mypy` recommends this way. https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#useful-built-in-types

But it's not my call.
@yegor256 what do you think on this? Raise a runtime error or instead assert and silence the checker for a particular rule?

test/patterns/incomplete_for/test_incomplete_for.py
ivanovmg: I see your point, but recently we had a discussion https://github.com/cqfn/aibolit/issues/776#issuecomment-2919820773 where @yegor256 highlighted that the java code does not necessarily have to compile for aibolit to check.
If the code is not syntactically correct, then we will get an error, when constructing an `AST`.

test/patterns/incomplete_for/test_incomplete_for.py
```diff
@@ -0,0 +1,67 @@
...
+
+import os
+from unittest import TestCase
+from pathlib import Path
+from aibolit.ast_framework.ast import AST
+from aibolit.patterns.incomplete_for.incomplete_for import IncompleteFor
+from aibolit.utils.ast_builder import build_ast
+
+
+class IncompleteForTestCase(TestCase):
```
ivanovmg, line 12: @AntonProkopyev I've just noticed this. Please consider that this project moves away from `unittest` and towards `pytest` https://github.com/cqfn/aibolit/issues/684. In future PRs you may want to build tests using `pytest` conventions.

ivanovmg:
@yegor256 it looks good to me.
Topics for discussion:
- https://github.com/cqfn/aibolit/pull/800#discussion_r2122490925
- https://github.com/cqfn/aibolit/pull/800#discussion_r2123463945

test/patterns/incomplete_for/test_incomplete_for.py
yegor256: @ivanovmg I believe that file system (similar to memory, threads, and sockets) is a resource that doesn't need to be mocke. Instead, we should expect _them_ to be fast enough for unit testing.

yegor256 (approved):

yegor256:
@rultor merge

yegor256:
@AntonProkopyev thanks!

0crat:
@antonprokopyev Thank you for your contribution! Your effort is appreciated. You've earned +24 points for this submission: +16 as a base reward, +8 for the 181 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) you wrote (capped at 16 points as per our policy). Great job on keeping your contribution concise and manageable. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) up the good work! Your current balance stands at [+21](https://cqfn.github.io/aibolit/aibolit-vitals.html).

Issue #813, opened 2025-06-07
Inconsistent Interface Contracts for Metrics/Patterns (AST vs. Filepath)


AntonProkopyev:
The current codebase exhibits inconsistencies in metric and pattern interfaces, where some implementations accept an `ast` while others take a `filepath`. Additionally, the unstructured use of dictionaries in `config` and `main` complicates maintenance and refactoring.  

Historical commit patterns suggest that the intended contract operates on an `ast` (Abstract Syntax Tree), but this is not formally enforced, leading to potential bugs and technical debt.

0pdd:
@AntonProkopyev 3 puzzles [#820](https://github.com/cqfn/aibolit/issues/820), [#821](https://github.com/cqfn/aibolit/issues/821), [#822](https://github.com/cqfn/aibolit/issues/822) are still not solved.

0pdd:
@AntonProkopyev 3 puzzles [#820](https://github.com/cqfn/aibolit/issues/820), [#829](https://github.com/cqfn/aibolit/issues/829), [#830](https://github.com/cqfn/aibolit/issues/830) are still not solved; solved: [#821](https://github.com/cqfn/aibolit/issues/821), [#822](https://github.com/cqfn/aibolit/issues/822).

0pdd:
@AntonProkopyev the puzzle [#829](https://github.com/cqfn/aibolit/issues/829) is still not solved; solved: [#820](https://github.com/cqfn/aibolit/issues/820), [#821](https://github.com/cqfn/aibolit/issues/821), [#822](https://github.com/cqfn/aibolit/issues/822), [#830](https://github.com/cqfn/aibolit/issues/830).

ivanovmg:
@AntonProkopyev I found that implementation of the desirable interface for some metrics, including, but not limited to `Entropy`, `LCOM4`, `IndentationCounter`, is not feasible.
The problem is that this code relies on the actual source code and/or `javalang` nodes, rather than `aibolit.ast.ASTNode` objects.
I guess, it is still possible to implement, but it is likely to go against the grain.
What do you think?

0crat:
[...]

PR #814, opened 2025-06-07
Enforce AST-based Pattern and Metric interface contracts

AntonProkopyev:
#813

> The current codebase exhibits inconsistencies in metric and pattern interfaces, where some implementations accept an `ast` while others take a `filepath`. Additionally, the unstructured use of dictionaries in `config` and `main` complicates maintenance and refactoring.
> 
> Historical commit patterns suggest that the intended contract operates on an `ast` (Abstract Syntax Tree), but this is not formally enforced, leading to potential bugs and technical debt.

This PR formalizes the expected interface contract by:

1. **Introducing `Mypy` type hints** to explicitly declare parameter and return types.
2. **Adding unit tests** with `inspect` to validate function signatures at runtime.

## Benefits

- **Clearer Contracts**: Explicit typing reduces ambiguity in interfaces.
- **Better Maintainability**: Static checks catch type-related errors early.
- **Future-Proofing**: Simplifies refactoring and enforces consistency for new contributions.

## Impact

- No breaking changes to existing functionality.
- Lays groundwork for future cleanup of `config`/`main` dictionary usage.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced static typing and type hinting for improved code clarity and type safety.
  - Added protocol and TypedDict classes to better define expected interfaces and configuration structures.
  - Added assertions to ensure metric objects conform to expected types.

- **Tests**
  - Replaced empty test class with new tests verifying type annotations and required parameters in configuration factories.
  - Added targeted tests for specific metrics and patterns, with clear instructions for future updates.

- **Style**
  - Applied type checking decorators to disable type checks where necessary for compatibility.

### DIFF
--- a/aibolit/config.py
+++ b/aibolit/config.py
@@ -59,6 +59,7 @@ from aibolit.patterns.hybrid_constructor.hybrid_constructor import HybridConstru
 from aibolit.patterns.var_decl_diff.var_decl_diff import VarDeclarationDistance as P20
 from aibolit.patterns.var_middle.var_middle import VarMiddle as P21
 from aibolit.patterns.var_siblings.var_siblings import VarSiblings as P27
+from aibolit.types_decl import LineNumber
 
 
 class Singleton(type):
@@ -74,7 +75,7 @@ class Singleton(type):
 
 @typing.runtime_checkable
 class Pattern(typing.Protocol):
-    def value(self, ast: AST) -> list[int]:
+    def value(self, ast: AST) -> list[LineNumber]:
         ...
 
 


aibolit/config.py
```diff
@@ -108,7 +142,7 @@ def test_csv():
         return os.environ.get('HOME_TEST_DATASET')
 
     @staticmethod
-    def get_patterns_config():
+    def get_patterns_config() -> PatternsConfig:  # ty: ignore[invalid-return-type]
```
AntonProkopyev: Ignore until https://github.com/astral-sh/ty/issues/154

AntonProkopyev:
@coderabbitai review

0crat:
@antonprokopyev Hello team member, I noticed that your branch name `improve-typings` doesn't follow our naming convention. As per our [policy](https://www.yegor256.com/2014/04/15/github-guidelines.html), branch names should match the ticket number you're working on. In this case, `813` would have been the correct name. Unfortunately, this results in a -12 point deduction. Your current balance is now [+45](https://cqfn.github.io/aibolit/aibolit-vitals.html). For future reference, please ensure your branch names align with our guidelines. Let's work together to maintain consistency in our project management. 👍

AntonProkopyev:
@yegor256 please take a look here

yegor256:
@ivanovmg what do you think?

aibolit/config.py
```diff
@@ -70,6 +72,38 @@ def __call__(cls, *args, **kwargs):
         return cls._instances[cls]
 
 
+@typing.runtime_checkable
+class Pattern(typing.Protocol):
+    def value(self, ast: AST) -> list[int]:
```
ivanovmg: @yegor256 @AntonProkopyev in my opinion, having to pass `ast` (or `filepath`) into `value` is not the best idea.
The reason for this is that it makes a class essentially a function.

``` python
class SomePattern:
  def __init__(self) -> None:
    pass  # why would we even need a class without an instance variable?

  def value(self, ast: AST) -> list[LineNumber]:
    # here is a pure function, which does not have any state (ast is not even an attribute)
```

Either we have a class with `ast` passed as a parameter:
``` python
class SomePattern:
  def __init__(self, ast: AST) -> None:
    self.ast = ast

  @classmethod
  def from_filepath(cls, filepath: str | os.PathLike) -> SomePattern:
    # an alternative constructor, which is easy to use for the end user
    ast = ...  # create ast from the filepath
    return cls(ast)

  def value(self) -> list[LineNumber]:
    ...
```

Or we have a simple function:
``` python
def some_pattern(ast: AST) -> list[LineNumber]:
  ...
```

I personally gravitate towards a class with `ast` as an input parameter, since it allows for an alternative constructor via a classmethod.
But still, a functional approach may work: there is a library `variants`, which enables function overloading https://github.com/python-variants/variants. However, the project hasn't been updated for quite a while.

Example:
``` python
matching_lines = some_pattern(ast)
matching_lines = some_pattern.from_filepath(filepath)
```

ivanovmg: If we are defining an interface, then may it be a good idea to explicitly type the return value as `list[LineNumber]`, where `LineNumber` is defined as a type explicitly:
``` python
LineNumber = NewType('LineNumber', int)
```
This way we will not mix up a simple integer with an integer, which represents a line number.

ivanovmg (commented):
@yegor256 @AntonProkopyev I think it is good idea that the interface get formalized. 
But I suppose we need to put more thought in this before we start making drastic changes to the whole code base.
Here are my thoughts on the `Pattern` interface in particular.

aibolit/config.py
AntonProkopyev: @ivanovmg (cc: @yegor256) thank you for review! I agree that the "function"-like interface is not a best idea. I think before any refactorings we need to provide clear and obvious contracts between modules and subsystems. After that making the code consistent. And then making it beautiful.

Let me show the steps of refactorings with pseudocode:

1. Initial state - unclear contracts, code will brakes
```python
class P1: # works
  def value(ast: AST) -> list[LineNumber]: ...
class P2: # wrong
  def value(filepath: str) -> list[LineNumber]: ...
```

2. Involve the contract which show me as a developer how all patterns should look,  all new patterns matches the contract because of tests:
```python
class Pattern(Protocol):
  def value(ast: AST) -> list[LineNumber]: ...

class P1: # works
  def value(ast: AST) -> list[LineNumber]: ...

class P2: # wrong # type: ignore
  def value(filepath: str) -> list[LineNumber]: ...
```

3. Fix the rest of patterns
```python
class P1(Pattern): # works
  def value(ast: AST) -> list[LineNumber]: ...

class P2(Pattern): # now works
  def value(ast: AST) -> list[LineNumber]: ...
```

4. After that maybe we realize that we need to provide not only ast, also a code, filename, etc and need get not only a line numbres, but also columns. We involving a new contract:
```python
class Source:
  def ast() -> AST: ...
  def code() -> str: ...
  def name() -> str: ...

class Findings:
  def locations() -> Iterable[Location]: ...

class PatternV2(Protocol):
  def findings_in(soruce: Source) -> Findings: ...

class P1(Pattern, PatternV2): # supports a new contract
  def value(ast: AST) -> list[LineNumber]: ...
  def findings_in(soruce: Source) -> Findings: ...

class P2(Pattern):
  def value(ast: AST) -> list[LineNumber]: ...
```

5. Support the new contract everywhere in code:
```python
class P1(Pattern, PatternV2): ...
class P2(Pattern, PatternV2): ...
```

6. Remove the obsolete contract:
```python
class P1(PatternV2): ...
class P2(PatternV2): ...
```

7. Finally rename `PatternV2` -> `Pattern`
```python
class P1(Pattern): ...
class P2(Pattern): ...
```

In conclusion with this approach we can incrementally fix the entire codebase with small PRs minimizing the risk of braking the code. The current PR is a second step of proposed approach. What do you think?

ivanovmg: @AntonProkopyev I appreciate the way you think of a potential future transition towards `Source` and `Findings`! This indeed can be a probable outcome of the project evolution, in my opinion. It is a good point, and I haven't thought of this.

My question still stands. What if we avoid passing `ast` and/or `filepath` into `value` method?
``` python
class Pattern(Protocol):
  def value(self) -> list[LineNumber]: ...

class ConcretePattern:
  def __init__(self, ast: AST) -> None:
    self.ast = ast

  def value(self) -> list[LineNumber]:
    return ...
```
I mean, `Pattern` is an object for a particular `AST`, right? Otherwise, it will become a class with static methods, and can be replaced with a function.
Or do you think of some parameters to be injected into `Pattern` instance apart from `ast`?
I can understand, that metric can have a tuning parameters, like (`max_limit`, or something, to throw an error when exceeded), but still even in this case making it be composed out of `AST` and some tuning parameters still seems a better option to me from the object thinking standpoint.

Then the sequence of the changes proposed from your side will be quite similar:
``` python
class Source:
  def ast() -> AST: ...
  def code() -> str: ...
  def name() -> str: ...

class Findings:
  def locations() -> Iterable[Location]: ...

class PatternV2(Protocol):
  def findings(self) -> Findings: ...

class P1(Pattern, PatternV2): # supports a new contract
  def value(self) -> list[LineNumber]: ...
  def findings(self) -> Findings: ...

class P2(Pattern):
  def value(self) -> list[LineNumber]: ...
```

@yegor256 what do you think from the object perspective regarding `Pattern` and `Metric`? Should they be composed out of `AST` or should they take `AST` as a parameter into instance methods (which will look more like static methods, IMHO)?

AntonProkopyev: @ivanovmg I agree, we can use a single `Pattern` instance per `AST` instance, and move an `ast` paramete to `__init__` method but this change requires patching of all patterns and control code - bloats the current PR.

Probably we should make another PR with chain of changes proposed - moving `ast` to constructor.

On the other hand there are patterns with hyperparameters, like `P20`, but we can bind a parameters with a lambda.

ivanovmg: @AntonProkopyev sure, let's proceed with your suggestion. I wonder how much we will have to change in future to adhere to the new interface though :)

AntonProkopyev:
@yegor256 сan you please merge the PR?

0crat:
@ivanovmg Great job on the code review! 🎉 You've earned +12 points, which is the base reward for reviewing someone else's contribution. Your running balance is now [+256](https://cqfn.github.io/aibolit/aibolit-vitals.html). Keep up the good work and remember that additional points can be earned based on the number of hits-of-code and comments in the review, as per our [work policy](https://example.com/work-policy).

0crat:
@antonprokopyev Thanks for your contribution! 🚀 You've earned a solid +19 points: +16 base, +7 for your 141 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and a small -4 deduction due to 20 review comments (our policy caps at 8). Keep up the great work and [focus on quality](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) to maximize your rewards. Your current balance stands at [+33](https://cqfn.github.io/aibolit/aibolit-vitals.html). Remember, contributions under 20 hits-of-code incur an 8-point deduction, so aim for impactful changes!

PR #818, opened 2025-06-12
Add a "quiet" exit

### LINKED ISSUE #567: Interrupt signal doesn't work
When I send interrupt signal with `Ctrl+C` CLI shows a lot of errors with stacktraces but doesn't stop.

alchemmist:
Close #567 

Added possibility of “quiet” exiting by pressing Ctrl+C. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of interrupt signals (e.g., Ctrl+C) to ensure the application responds as expected during concurrent operations.
  - Enhanced graceful shutdown to prevent error tracebacks from appearing when interrupting the application.  
- **Tests**
  - Added a test to verify the application launches and shuts down gracefully upon receiving an interrupt signal without error messages.


test/interface/test_graceful_shutdown.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
+# SPDX-License-Identifier: MIT
+import subprocess
+import time
+import signal
+
+
+def test_quiet_exiting():
+    proc = subprocess.Popen(
+        ["python", "aibolit"],
```
ivanovmg: The bot suggested that `"python"` is replaced with `sys.executable` to ensure reproducibility in various environments.

```suggestion
        [sys.executable, "aibolit"],
```
@alchemmist can you update this PR?
And maybe merge master (rebase on top of master).

alchemmist: @ivanovmg, thanks for review!

Yes, I'll do it. Below sent the commits with the sync with `master` and the fixed test.

ivanovmg (approved):
@alchemmist sorry for the delay!
Looks good to me.
@yegor256

yegor256 (approved):

alchemmist:
@ivanovmg, does you have any ideas, why on CI and locally tests completed, bot in rultor failed?

test/interface/test_graceful_shutdown.py
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: MIT
+import sys
+import subprocess
+import time
+import signal
+
+
+def test_quiet_exiting():
+    proc = subprocess.Popen(
+        [sys.executable, 'aibolit'],
```
ivanovmg: Since the error emitted by `rultor` was related to `runpy`, I assume that maybe running this explicitly as a module may help:
```suggestion
        [sys.executable, '-m', 'aibolit'],
```

yegor256:
@rultor merge

test/interface/test_graceful_shutdown.py
```diff
@@ -0,0 +1,48 @@
...
+    process = subprocess.Popen(
+        [sys.executable, '-m', 'aibolit', 'check', '--folder', 'test'],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+        bufsize=1,
+        universal_newlines=True,
+    )
+
+    time.sleep(3)
```
ivanovmg, line 19: It becomes a really slow test, I suppose.
I guess you take this long timeout to ensure that `get_version` successfully passes.
But this is flaky and also makes tests to run way longer.

In my opinion, we can probably avoid it by thinking in this direction:
``` python
def setup_graceful_exiting():
  # implement signal handlers here
  # this function should be tested
  ...

if __name__ == '__main__':
  setup_gracefully_exiting()
  main()
```

Then the function `setup_graceful_exiting` can be tested separately without running long processes - can simply run `[sys.executable,  '-v']`, I guess.

Useful read, IMHO https://davidhamann.de/2022/09/29/handling-signals-in-python/

aibolit/__main__.py
```diff
@@ -906,6 +910,8 @@ def main():
             requests.exceptions.ReadTimeout,
     ):
         print("Can't check aibolit version. Network is not available or PyPI does not respond")
+    except KeyboardInterrupt:
+        sys.exit(0)
```
ivanovmg, line 914: It is not clear to my why handling `get_versions` is separate from the rest of the `main` function.
Probably, this is done to make sure tests pass, but there must be a better way, I presume.

ivanovmg (commented):
@alchemmist please have a look at my comments. We may want to extract signal handling functionality in a separate function to avoid flaky tests and make them fast.
Correct me if I'm wrong in my observations.

aibolit/__main__.py
```diff
@@ -879,6 +880,8 @@ def run_thread(files, args):
     :param files: list of java files to analyze
 
     """
+    signal.signal(signal.SIGINT, signal.SIG_DFL)
```
kreinba, line 883: Setting SIG_DFL here defeats the stated goal. SIG_DFL on SIGINT raises KeyboardInterrupt in the main thread, which then propagates past `except Exception` in main (KeyboardInterrupt does not subclass Exception) and prints a traceback. The only KeyboardInterrupt handler in this file sits around get_versions, which has already returned by the time run_thread runs. Either install a custom handler that calls `sys.exit(0)`, or wrap the second try in main with `except KeyboardInterrupt: sys.exit(0)` so Ctrl+C during analysis exits cleanly.

aibolit/__main__.py
```diff
@@ -895,6 +898,7 @@ def get_versions(pkg_name):
 
 
 def main():
+    print('aibolit has been launched', flush=True)
```
kreinba, line 901: An unconditional `print('aibolit has been launched', ...)` on every invocation pollutes stdout for every caller (pipelines, CI, scripts that parse the output). The line exists only so the new test can detect startup; it should not ship as a permanent user-visible change. Drop the print and replace the test's stdout poll with a small sleep that gives the version check time to finish, or gate the print behind a verbose flag.

aibolit/__main__.py
```diff
@@ -9,6 +9,7 @@
 import concurrent.futures
 import json
 import multiprocessing
+import signal
```
kreinba, line 12: `import signal` is placed between `multiprocessing` and `operator`, which breaks the alphabetical ordering of the stdlib import block. It should sit between `pickle` and `sys`. Likely to be flagged by isort/flake8 on a fresh sort.

kreinba (changes requested):
The quiet-exit goal is not delivered. run_thread resets SIGINT to SIG_DFL, so a Ctrl+C during the actual analysis raises KeyboardInterrupt that nothing catches: the second try block in main wraps it as except Exception, and KeyboardInterrupt does not inherit from Exception. The new test only sends SIGINT during the get_versions window, which is the one place a KeyboardInterrupt handler exists, so it never exercises the path the PR claims to fix. Three inline comments below.

aibolit/__main__.py
bibonix, line 901: This `print` is written to stdout for every subcommand, including `check --format xml`, where line 785 writes the XML tree to `stdout.buffer`. The launch banner now precedes the XML, so any consumer parsing the documented XML output of `aibolit check --format xml ...` will fail to parse stdout as XML. The `version` subcommand and any other machine-readable consumer are affected the same way. If a sentinel is needed only so the new test can detect startup, write it to stderr (`print(..., file=sys.stderr, flush=True)`) or remove it entirely and have the test wait on a condition that does not require modifying production output.

test/interface/test_graceful_shutdown.py
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
```
bibonix, line 1: `test/interface/` is a new package directory but lacks `__init__.py`, while every sibling under `test/` (`test/patterns/`, `test/config/`, `test/integration/`, `test/ast_framework/`, `test/recommend/`, `test/utils/`, `test/metrics/`, `test/stats/`) ships one. Please add an empty `test/interface/__init__.py` with the standard SPDX header to keep the test layout consistent and avoid surprises with rootdir/conftest discovery when the test is moved or invoked outside the Makefile's `python3 -m pytest test/` entry point.

bibonix (changes requested):
Requesting changes. The strongest concrete defect not yet flagged inline is that the unconditional launch print corrupts the documented `--format xml` output contract by emitting plaintext before the XML tree on stdout. Two inline comments below.

Issue #696, opened 2025-05-21
Unit tests performance issue due to inefficient minimal working example extraction


ivanovmg:
There is a bunch of slow unit tests.

``` python
$ python -m coverage run -m unittest discover --durations 20
...

Slowest test durations
----------------------------------------------------------------------
44.788s    test_ncss (test.ast_framework.test_java_class_decomposition.JavaClassDecompositionTestSuite.test_ncss)
3.119s     test_more_method_invocations (test.patterns.send_null.test_send_null.SendNullTestCase.test_more_method_invocations)
3.042s     test_another_setter_patterns (test.patterns.classic_setter.test_classic_setter.SetterTestCase.test_another_setter_patterns)
3.029s     test_three_writers_one_reader (test.patterns.er_class.test_er_class.ErClassTestCase.test_three_writers_one_reader)
3.019s     test_implements_many (test.patterns.implements_multi.test_implements_multi.ImplementsMultiTestCase.test_implements_many)
2.910s     test_four_setter_patterns (test.patterns.classic_setter.test_classic_setter.SetterTestCase.test_four_setter_patterns)
2.900s     test_multi_level_invocation (test.patterns.send_null.test_send_null.SendNullTestCase.test_multi_level_invocation)
2.797s     test_class_parser (test.patterns.er_class.test_er_class.ErClassTestCase.test_class_parser)
2.769s     test_implements_with_nested_parantheses (test.patterns.implements_multi.test_implements_multi.ImplementsMultiTestCase.test_implements_with_nested_parantheses)
2.072s     test_many_chains (test.patterns.method_chaining.test_method_chaining.MethodChainTestCase.test_many_chains)
1.615s     test_bad_class2 (test.patterns.var_decl_diff.test_var_decl_diff.VarDeclarationDiffTestCase.test_bad_class2)
1.507s     test_string_with_quotes (test.patterns.string_concat.test_concat_string.ConcatStringTestCase.test_string_with_quotes)
1.380s     test_catch_with_similar_name (test.patterns.redundant_catch.test_redundant_catch.RedundantCatchTestCase.test_catch_with_similar_name)
1.279s     test_calculate_patterns_and_metrics (test.recommend.test_recommend_pipeline.TestRecommendPipeline.test_calculate_patterns_and_metrics)
1.225s     test_calculate_patterns_and_metrics_with_suppress (test.recommend.test_recommend_pipeline.TestRecommendPipeline.test_calculate_patterns_and_metrics_with_suppress)
[...]

PR #828, opened 2025-06-13
Use pytest-testmon to speed up tests in local development

AntonProkopyev:
Improve performance between test runs. With `--testmon` the only affected by code changes tests will be executed.

```shell
time make test
python3 -m pytest --testmon --cov=aibolit/ test/
...
real    0m1.773s
user    0m4.351s
sys     0m0.193s
```

See: https://testmon.org/

#696

## Summary by CodeRabbit

- **Chores**
  - Updated `.gitignore` to exclude files related to test monitoring.
  - Added `pytest-testmon` as a new dependency.
  - Modified test commands to include test monitoring functionality.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ install:
 	python3 aibolit --version
 
 test:
-	python3 -m pytest --cov=aibolit/ test/
+	python3 -m pytest --testmon --cov=aibolit/ test/
 
 it:
 	python3 -m test.integration.test_patterns_and_metrics
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@
 .Python
 .ropeproject
 .ruff_cache/
+.testmondata*
 .tox/
 .venv
 .vscode/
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ install:
 	python3 aibolit --version
 
 test:
-	python3 -m pytest --cov=aibolit/ test/
+	python3 -m pytest --testmon --cov=aibolit/ test/
```
ivanovmg, line 24: I believe that using this plugin is reasonable when running in a development mode, to save time when running a test suite.

I wonder if running the same in the workflow will still execute the whole test suite. It is essential to run all tests in the workflow to make every run exactly the same and fully reproducible.

AntonProkopyev: @ivanovmg Thank you for the comment!

If tests in the workflow running in the fresh environment and there is no `.testmondata` file then all tests will be runned.

ivanovmg (approved):
Looks good to me.
@yegor256

yegor256 (approved):

0crat:
@yegor256 Thanks for the review! You've earned +4 points for this: +12 as a basis; -10 for very few (2) comments; -8 for too few (4) hits-of-code; +10 to give you at least something.  Your running balance is [-160](https://cqfn.github.io/aibolit/aibolit-vitals.html).

Issue #742, opened 2025-05-27
bidirect_index.py:18-25: Implement bidirect index pattern...


0pdd:
The puzzle `139-658fe398` from #139 has to be resolved: 

https://github.com/cqfn/aibolit/blob/cc472ae91f3b2d6f7b87f5d220ab1bc6265b2865/aibolit/patterns/bidirect_index/bidirect_index.py#L18-L25

The puzzle was created by @yegor256 on 27-May-25. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #842, opened 2025-06-30
bidirect_index.py:18-25: Implement bidirect index pattern

KachanovYev:
https://github.com/cqfn/aibolit/issues/742

Implement bidirect index pattern
tests is working
add docs for code

## Summary by CodeRabbit

* **New Features**
  * Added detection of bidirectional index variables in Java source files, identifying variables both incremented and decremented within the same method or block, excluding increments/decrements inside for-loops with local variables of the same name.
  * Extended AST node types to include augmented assignments, identifiers, and unary operations.

* **Documentation**
  * Added detailed explanations describing the bidirectional index detection concept and usage.

* **Tests**
  * Removed unused imports and decorators from bidirectional index pattern tests for cleaner test code.

### DIFF
--- a/aibolit/patterns/bidirect_index/bidirect_index.py
+++ b/aibolit/patterns/bidirect_index/bidirect_index.py
@@ -54,7 +54,7 @@ class BidirectIndex:
         # Regular expression to find method declaration in Java
         method_pattern = re.compile(r'''
             (public|private|protected|static|\s)*  # access modifiers and static
-            ([\w<>\[\]]+)                          # # return type
+            ([\w<>\[\]]+)                          # return type
             \s+                                    # space
             \w+                                    # method name
             \s*                                    # possible spaces
@@ -92,7 +92,7 @@ class BidirectIndex:
                 j = i + 1
                 while j <= end:
                     line_ = lines[j]
-                    # Check for re-declaration of variable with type
+                    # Check for both typed declarations and simple reassignments
                     if re.match(
                             r'\s*(int|long|byte|short)\s+' + re.escape(var) + r'\s*=', line_):
                         break


aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,153 @@
...
+    @staticmethod
+    def find_for_blocks(lines, var, start, end):
+        """
+        Find all ranges of for-loops that declare a local variable with the given name.
+        """
+        for_blocks = []
+        k = start
+        while k < end:
+            l = lines[k]
+            for_decl = re.match(r'\s*for\s*\(\s*int\s+' + re.escape(var) + r'\s*=', l)
```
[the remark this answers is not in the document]
KachanovYev: resolved

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,153 @@
...
+            line = lines[i]
+            typed_decl = re.match(r'\s*(int|long|byte|short)\s+(\w+)\s*=', line)
+            untyped_decl = re.match(r'^\s*(\w+)\s*=\s*[^;]+;', line) \
+                if not typed_decl else None
+            var = typed_decl.group(2) if typed_decl else (untyped_decl.group(1)
+                                                          if untyped_decl else None)
+            if var:
+                j = i + 1
+                while j <= end:
+                    l = lines[j]
```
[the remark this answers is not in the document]
KachanovYev: resolved

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,153 @@
...
+        Count increments and decrements of variable `var` outside any for-blocks.
         """
-        return []
+        inc_outside = 0
+        dec_outside = 0
+        k = start
+        while k < end:
+            in_for = any(bstart <= k <= bend for (bstart, bend) in for_blocks)
+            if not in_for:
+                l = lines[k]
```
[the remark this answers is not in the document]
KachanovYev: resolved

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,164 @@
...
 
 class BidirectIndex:
+    """
+    This class analyzes Java source code to find line numbers where a variable
+    is used as a bidirectional index.
+    A bidirectional index is defined as a variable that:
+        - is assigned a value (with or without type declaration) within a method or block,
+        - is incremented and decremented somewhere in its scope,
+        - increments/decrements inside a for-loop with a local variable of the same name
+        are ignored (to avoid "fake" cases).
```
ivanovmg: Maybe indent it to match the indentation of the corresponding item

```suggestion
        - increments/decrements inside a for-loop with a local variable of the same name
          are ignored (to avoid "fake" cases).
```

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,164 @@
...
+        result = idx.value('MyClass.java')
+        # result is a list of line numbers matching the described pattern
+    """
 
     def __init__(self):
         pass
 
-    def value(self, filename: str | os.PathLike):
+    @staticmethod
+    def value(filename: str | os.PathLike):
```
ivanovmg: Please follow the [interface](https://github.com/cqfn/aibolit/pull/814/files#diff-2463d356cca43b17be5e04f92ab65b6de962cd845b9712c85b5a1246784be97dR76-R79) described here: https://github.com/cqfn/aibolit/pull/814 (related to https://github.com/cqfn/aibolit/issues/813).

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,164 @@
...
+            with open(filename, encoding='utf-8') as f:
+                lines = f.readlines()
+        except (FileNotFoundError, IOError) as e:
+            raise ValueError(f'Failed to read file {filename}: {e}')
 
-        :param filename: filename to be analyzed
-        :return: list of LineNumber with the variable declaration lines
+        result: list[int] = []
+        for m_start, m_end in BidirectIndex.find_methods(lines):
+            BidirectIndex.analyze_block(lines, m_start + 1, m_end, result)
```
ivanovmg: Can we maybe use instance methods, rather than static methods?

```suggestion
        for m_start, m_end in self.find_methods(lines):
            self.analyze_block(lines, m_start + 1, m_end, result)
```

Or, if reasonable, we can extract functions, that detect the pattern, but make `BidirectIndex.value` simply call this function.

aibolit/patterns/bidirect_index/bidirect_index.py
ivanovmg: Oh, I see, in this implementation you totally relied on the literate source code of the file, rather than the abstract syntax tree.
Although, your solution covers all test cases, it goes against the direction on the unification of the `Pattern` interface.

I suppose that if this is a working solution, then having resolved the comments, it may be merged **with a puzzle** to refactor it into using an `ast_framework.AST` rather than a `filename` as an input parameter.
@AntonProkopyev can you suggest the way forward here?

ivanovmg: I suggest against use of a static method, at least in this particular case, where it is a public method.
```suggestion
    def value(self, filename: str | os.PathLike):
```

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,164 @@
...
+        Returns:
+            list[tuple[int, int]]: Each tuple is (start_line_idx, end_line_idx) for a method.
+        """
+        res = []
+        brace = 0
+        mstart = None
+        for idx, line in enumerate(lines):
+            if (re.search(
+                    r'(public|private|protected|static|\s)*([\w<>\[\]]+)\s+\w+\s*\([^)]*\)\s*{',
+                    line)):
```
ivanovmg: I suppose this deserves a separate function/method.
Maybe extract a function:
``` python
# or whatever the name should be
# I can't tell from the regex what it actually does :)
def _is_index_definition(line: str) -> bool:
  ...
```

ivanovmg: These long regexes are really hard to read. But the good news is that they can be made very readable using `re.VERBOSE` flag.
Please have a look at the examples here: https://pycon2017.regex.training/re-module#verbose-flag

Example in action:
image

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,164 @@
...
+                while k < end and brace > 0:
+                    brace += lines[k].count('{') - lines[k].count('}')
+                    k += 1
+                for_blocks.append((bstart, k - 1))
+                continue
+            k += 1
+        return for_blocks
+
+    @staticmethod
+    def count_inc_dec(lines, var, for_blocks, start, end):
```
ivanovmg: I suggest that all methods apart from `value` are made internal by adding a leading underscore. Or, if you gravitate towards static methods, then better extract functions, outside the class.

aibolit/patterns/bidirect_index/bidirect_index.py
ivanovmg: @KachanovYev can you please follow up on this? These regexes are still too difficult to digest unless properly commented like described here.

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,157 @@
...
+        Returns:
+            List[int]: Sorted list of line numbers where bidirectional indices are found.
+        """
+        result: list[int] = []
+        for m_start, m_end in BidirectIndex.find_methods(lines):
+            BidirectIndex.analyze_block(lines, m_start + 1, m_end, result)
+        return sorted(result)
+
+    @staticmethod
+    def find_methods(lines):
```
ivanovmg: Please consider removing `staticmethod` everywhere, or extracting a related functionality into separate functions (https://github.com/cqfn/aibolit/pull/842/files#r2176314046).
```suggestion
    def _find_methods(self, lines):
```

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,157 @@
...
+        return for_blocks
 
-        @todo #139:30min Implement bidirect index pattern
-        If the same numeric variable is incremented and decremented within the same method,
-        it's a pattern. A numeric variable should either be always growing or decreasing.
-        Bi-directional index is confusing. The method must return a list with the line numbers
-        of the variables that match this pattern. After implementation, activate tests in
-        test_bidirect_index.py
+    @staticmethod
+    def count_inc_dec(lines, var, for_blocks, start, end):
```
ivanovmg: https://github.com/cqfn/aibolit/pull/842/files#r2176330476

```suggestion
    def _count_inc_dec(self, lines, var, for_blocks, start, end):
```

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,157 @@
...
+        lines = open("MyClass.java", encoding="utf-8").readlines()
+        result = idx.value(lines)
+        # result is a list of line numbers matching the described pattern
+    """
 
     def __init__(self):
         pass
 
-    def value(self, filename: str | os.PathLike):
+    def value(self, lines):
```
ivanovmg: I don't think that adding a third kind of interface for `value` is a good idea.

This is what we already have:
``` python
def value(self, ast: AST) -> list[LineNumber]: ...  # preferable
def value(self, filename: str | os.PathLike) -> list[LineNumber]: ...  # soon to be deprecated in favor of the preferable option above
```

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,154 @@
...
+                j = i + 1
+                while j <= end:
+                    line_ = lines[j]
+                    # Check for both typed declarations and simple reassignments
+                    if re.match(r'\s*(int|long|byte|short)\s+' + re.escape(var) + r'\s*=', line_):
+                        break
+                    j += 1
+                inc_outside, dec_outside = (
+                    self._count_inc_dec(lines, var, i + 1, j))
+                if inc_outside and dec_outside:
```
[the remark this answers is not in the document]
KachanovYev: resolved

KachanovYev:
@ivanovmg 
Hi PR is updated by your request.

test/patterns/bidirect_index/test_bidirect_index.py
```diff
@@ -1,69 +1,68 @@
...
 
-@unittest.skip('Not implemented')
 class BidirectIndexTestCase(TestCase):
     dir_path = Path(os.path.realpath(__file__)).parent
 
     def test_bidirect_index_increase_decrease(self):
+        file_path = Path(self.dir_path, 'BidirectIndexIncreaseDecrease.java')
         self.assertEqual(
-            BidirectIndex().value(Path(self.dir_path, 'BidirectIndexIncreaseDecrease.java')),
+            BidirectIndex().value(file_path),
```
ivanovmg: I suggest that we leave the changes in the test files outside of the present PR (apparently, except making the test non-skipped). It is OK to do it in another one, dedicated to the refactoring if necessary.

aibolit/patterns/bidirect_index/bidirect_index.py
```diff
@@ -1,25 +1,154 @@
...
+                    inc_outside += 1
+                if re.search(
+                        r'(--' + re.escape(var) + r'|' + re.escape(var) + r'--|' +
+                        re.escape(var) + r'\s*-=\s*1\b|' +
+                        re.escape(var) + r'\s*=\s*' + re.escape(var) + r'\s*-\s*1\b)', line_):
+                    dec_outside += 1
+            k += 1
+        return inc_outside, dec_outside
+
+    def value(self, filename: str | os.PathLike) -> list[LineNumber]:
```
ivanovmg: It is better to keep the public method on top, while having the remaining helper methods defined below. This way we will read the code like the newspaper - header first, then details, then more details, etc.

aibolit/patterns/bidirect_index/bidirect_index.py
ivanovmg: @KachanovYev I strongly encourage you to address this issue. Reading complex regexes in the future is difficult and reduces the level of maintainability.

See https://github.com/cqfn/aibolit/pull/842#discussion_r2176329295

ivanovmg (changes requested):
@KachanovYev please have a look at my review.

1. Could you move `value` to the top of the class, so that it is the public method, that is visible first to the reader?
2. Keep only essential changes in the test files, while moving those desirable into another PR.
3. Comment regexes, because it is extremely difficult to understand them. I suggest using `re.VERBOSE` flag

While issues 1 and 2 are not that critical, issue 3 is pretty concerning. Could you address it?

test/patterns/bidirect_index/test_bidirect_index.py
KachanovYev: resolved

aibolit/patterns/bidirect_index/bidirect_index.py
KachanovYev: resolved

PR #843, opened 2025-06-30
refactor: remove unnecessary getters

CAN4red:
* refactor: remove unnecessary getters and make fields public

* refactor: encapsulate all the fields making them private

## Summary by CodeRabbit

* **Refactor**
  * Removed several public methods for accessing image asset properties, reducing direct external access to these details.

### DIFF
--- a/test/ast_framework/LottieImageAsset.java
+++ b/test/ast_framework/LottieImageAsset.java
@@ -33,26 +33,6 @@ public class LottieImageAsset {
     this.dirName = dirName;
   }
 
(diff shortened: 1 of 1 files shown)


0crat:
@can4red It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`master`". You've earned -12 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`842`".

PR #861, opened 2025-07-09
Extend `MvnFreeNPathMetric` to handle `while` loops

### LINKED ISSUE #852: `MvnFreeNPathMetric` metric does not cover while loops
There is a development towards removal of `mvn` dependency from the package. As of now, we still have `NPathMetric`, which depends on `mvn`.
`MvnFreeNPathMetric` intends to replace `NPathMetric` once completed.

However, currently, `MvnFreeNPathMetric` does not support `while` loops.

It is necessary to extend test suite and add the corresponding implementation.

Xref:
- https://github.com/cqfn/aibolit/issues/796
- https://github.com/cqfn/aibolit/pull/849#issuecomment-3043797091

ivanovmg:
In this PR we extend `MvnFreeNPathMetric` to handle `while` loops.
All test cases are checked against `checkstyle`.

Closes #852


aibolit/metrics/npath/main.py
```diff
@@ -103,16 +103,15 @@ def _node_npath(self, node: ASTNode) -> int:
...
-            return self._sequence_npath(node.children)
+        dispatcher = {
+            ASTNodeType.IF_STATEMENT: self._if_npath,
+            ASTNodeType.SWITCH_STATEMENT: self._switch_npath,
+            ASTNodeType.FOR_STATEMENT: self._for_loop_npath,
+            ASTNodeType.WHILE_STATEMENT: self._while_loop_npath,
+            ASTNodeType.BINARY_OPERATION: self._binary_expression_npath,
+        }
+        handler = dispatcher.get(node.node_type, lambda n: self._sequence_npath(n.children))
+        return handler(node)
```
ivanovmg, line 118: Here I had to perform refactoring towards this dictionary dispatch pattern to avoid too many returns.
In particular, the linter was complaining about 7 return statements with a max of 6.
```
aibolit/metrics/npath/main.py:102:4: R0911: Too many return statements (7/6) (too-many-return-statements)
```

We can consider reverting and silencing the warning if the original implementation is considered more readable.

Alternatively, we can simply write resulting value to a variable `result` and then only return `result` once.

ivanovmg:
@AntonProkopyev could you review, please?

ivanovmg:
> @AntonProkopyev could you review, please?

@yegor256 FYI

aibolit/metrics/npath/main.py
```diff
@@ -136,11 +135,16 @@ def _group_npath(case_group: ASTNode) -> int:
...
-        if hasattr(node.control, 'condition') and node.control.condition:
-            condition_npath = max(1, self._node_npath(node.control.condition))
-        else:
-            condition_npath = 1
-        return body_npath + condition_npath
+        return body_npath + self._condition_npath(node.control)
+
+    def _while_loop_npath(self, node: ASTNode) -> int:
+        body_npath = self._node_npath(node.body)
+        return body_npath + self._condition_npath(node)
```
AntonProkopyev: @ivanovmg According to [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) I suggest to calculate npath for `condition` before `body` part.

test/metrics/npath/test_all_types.py
```diff
@@ -433,6 +433,110 @@ class Test {
         ''').strip()
         assert self._value(content) == 5
 
+    def test_simple_while_loops(self) -> None:
```
AntonProkopyev, line 438: @ivanovmg The tests are clear and sufficient! I would also suggest making sure that the tests from the previous implementation work for the new implementation.

aibolit/metrics/npath/main.py
AntonProkopyev: @ivanovmg IMO the dictionary dispatching is pretty clean and easy to read. Maybe we could consider add explicit `if` when default handle is needed but the `lambda` is also OK for me.

AntonProkopyev (commented):
@ivanovmg The PR is easy to read and understand. I have some minor considerations but the PR is well enough to get merged.

aibolit/metrics/npath/main.py
ivanovmg: @AntonProkopyev I extracted a separate function `default_handler`. I hope it helps convey the message better than an anonymous function. Does it look OK?

aibolit/metrics/npath/main.py
ivanovmg: @AntonProkopyev I moved the statements around to first calculate `condition_npath` and only then `body_npath`. Is it what you meant?

test/metrics/npath/test_all_types.py
ivanovmg: @AntonProkopyev the tests for the existing case files do not actually pass for `MvnFreeNPathMetric` as of now. I added the corresponding puzzles. Need some expertise in NPath complexity to fix the issues, I suppose. I tried to figure those out, but failed miserably.

0crat:
@ivanovmg Hey there! 👋 Great job on your contribution! You've earned +4 points for this one. Here's the breakdown: +16 as a base, +10.55 for your 211 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (though we had to deduct 8 for going over 200), and -16 for no code review. We bumped it up to 4 to keep you motivated. Remember, code reviews are crucial, and [keeping contributions smaller](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) can boost your points next time. Your balance is now [+115](https://cqfn.github.io/aibolit/aibolit-vitals.html). Keep the code coming!

PR #881, opened 2025-07-20
Improve error handling in run_thread

MMenshikh:
This PR improves error reporting in the run_thread function. Now any exceptions during Java file analysis are printed to the console instead of silently failing.

This helps developers debug pattern failures or malformed Java files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during concurrent tasks to prevent interruptions from individual errors.

* **Refactor**
  * Updated result processing to return a complete list of outcomes instead of delivering them one by one.

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -877,13 +877,16 @@ def run_thread(files, args):
     """
     Parallel patterns/metrics calculation
     :param files: list of java files to analyze
-
     """
+    results = []
     with concurrent.futures.ThreadPoolExecutor(max_workers=multiprocessing.cpu_count()) as executor:
         future_results = [executor.submit(run_recommend_for_file, file, args) for file in files]
-        concurrent.futures.wait(future_results)
-        for future in future_results:
-            yield future.result()
+        for future in concurrent.futures.as_completed(future_results):
+            try:
+                results.append(future.result())
+            except Exception as e:
+                print(f"Error analyzing file: {e}")
+    return results
 
 
 def get_versions(pkg_name):
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -885,7 +885,7 @@ def run_thread(files, args):
             try:
                 results.append(future.result())
             except Exception as e:
-                print(f"Error analyzing file: {e}")
+                print(f'Error analyzing file: {e}')
     return results
 
 


0crat:
@mmenshikh Hey there! 👋 I noticed your branch name `improve-error-handling` doesn't quite follow our naming convention. [It's not ideal](https://www.yegor256.com/2014/04/15/github-guidelines.html) and unfortunately, that means a -12 point penalty. 😕 Next time, try using the ticket number instead - something like `880` would be perfect! It helps keep things organized and aligns with our policy. Let's work on this together to avoid future penalties and keep our workflow smooth. 🚀

bibonix (commented):
test

aibolit/__main__.py
```diff
@@ -877,13 +877,16 @@ def run_thread(files, args):
...
     :param files: list of java files to analyze
-
     """
+    results = []
     with concurrent.futures.ThreadPoolExecutor(max_workers=multiprocessing.cpu_count()) as executor:
         future_results = [executor.submit(run_recommend_for_file, file, args) for file in files]
-        concurrent.futures.wait(future_results)
-        for future in future_results:
-            yield future.result()
+        for future in concurrent.futures.as_completed(future_results):
```
bibonix, line 884: `as_completed(future_results)` drops the future-to-file mapping that the old `for future in future_results` loop preserved. When `future.result()` raises, the except branch below has no way to recover which input file caused the failure. Build a `dict` keyed by future when submitting (`{executor.submit(run_recommend_for_file, file, args): file for file in files}`) and use it inside the except to identify the file.

aibolit/__main__.py
```diff
@@ -877,13 +877,16 @@ def run_thread(files, args):
...
+            try:
+                results.append(future.result())
+            except Exception as e:
```
bibonix, line 887: A bare `except Exception` swallows every error class, including programming bugs such as `AttributeError` or `KeyError` raised from inside the per-file analyzer. The PR title says "improve error handling", but catching `Exception` here hides regressions that should fail the run. Narrow the catch to the exception classes `run_recommend_for_file` is documented to raise (or define a project-specific `AnalysisError` and catch only that).

aibolit/__main__.py
```diff
@@ -877,13 +877,16 @@ def run_thread(files, args):
...
+                print(f'Error analyzing file: {e}')
```
bibonix, line 888: Two problems with this print. First, the message names neither the file that failed (see the inline comment on line 884) nor the exception type, so the operator cannot reproduce or triage. Second, error diagnostics should go to `sys.stderr`, not stdout; stdout in this tool is used for the XML/JSON report stream, and mixing error text into it will corrupt downstream pipes like `aibolit check | jq ...`. Use `print(f'Failed to analyze {file}: {type(e).__name__}: {e}', file=sys.stderr)`.

aibolit/__main__.py
```diff
@@ -877,13 +877,16 @@ def run_thread(files, args):
...
+    return results
```
bibonix, line 889: `run_thread` was a generator; this turns it into a function returning a list. The caller at line 777 (`results = list(run_thread(files, args))`) still works, but the contract change is invisible from the signature and the docstring two lines above still says "Parallel patterns/metrics calculation" without noting that all futures are now drained before returning. Either keep it as a generator and put the try/except around the inner `yield future.result()`, or update the docstring to state the eager semantics and add a return-type annotation.

bibonix (changes requested):
The change replaces a streaming generator with an eager list and adds a bare-except around `future.result()`. The intent is right, but the implementation drops information that callers need on failure and silently changes the function's contract. Four inline comments name the defects.

PR #897, opened 2025-09-16
#896: Replace `ty` with `pyright`

### LINKED ISSUE #896: `ty` is pre-release software and not ready for production use
Let's use `pyright` instead of `ty` until `ty` becomes production-ready

literally-bug-creator:
Resolve #896


yegor256:
@literally-bug-creator why not keeping them both in the CI pipeline? Also, in order to merge this PR we should fix all issues detected by pyright. Many thanks for this contribution!

literally-bug-creator:
@yegor256 I think it's better not to keep two type checkers, because there is a possibility of conflicts, and also `ty` doesn't find any errors in the current configuration, even though there are many of them

yegor256:
@literally-bug-creator I suggest we keep them both and wait for the first conflict to arrive. Then, we will decide which one is right (ty or pyright) and do something about it. WDYT?

literally-bug-creator:
I will continue the task in another PR

PR #904, opened 2025-09-28
Fix command recommend fail

kodsurfer:
## Summary by CodeRabbit

- Bug Fixes
  - Improved stability when processing analysis results by skipping unexpected items in pattern lists to prevent errors.

- Tests
  - Added unit tests to confirm XML report generation for both nested and flat result structures, ensuring pattern codes appear as expected.

- Chores
  - Added a test package initializer to organize XML-related tests.


aibolit/__main__.py
```diff
@@ -534,9 +534,9 @@ def _process_file_result(file, result_for_file):
             results = flatten(results)
         patterns_number = len(results)
         importance_for_class = []
-
+
```
kreinba, line 550: Stray trailing space on this otherwise blank line is what trips flake8 W293 and ruff. Restore the line to a clean blank.

test/xml/test_xml.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
+# SPDX-License-Identifier: MIT
+
+def test_xml_report_with_nested_results(self):
```
kreinba, line 4: These are top-level functions, not methods, but they take a `self` parameter and call `self.__create_mock_cmd()` and `self.assertIn(...)`. Wrap both in a `unittest.TestCase` subclass (`class TestXmlReport(unittest.TestCase): ...`) and move the helper `_create_mock_cmd` onto the class, otherwise pytest never finds them and unittest can't execute them.

kreinba, line 4: `create_xml_tree` is referenced but never imported. Add the missing `from aibolit.__main__ import create_xml_tree` (or the actual module it lives in) at the top of the file.

kreinba (changes requested):
The fix in __main__.py is reasonable, but the new test file does not work: the two test functions take a `self` parameter without being inside a TestCase class, and they reference `create_xml_tree` and `__create_mock_cmd` without importing or defining them. As written, the test module will not even import. CI is also red on flake8, pylint, ruff, and ty — at least one cause is the trailing whitespace introduced on the previously blank line. See inline comments for the specific fixes.

PR #959, opened 2025-10-15
Add `style` workflow

### LINKED ISSUE #955: There are multiple CI workflows just for style checking
We need to replace `ruff`, `pylint` and others with a lint scenario from our `Makefile`, which runs all of them

deyna256:
Resolve #955

## Summary by CodeRabbit

* **Chores**
  * Consolidated code-style and lint checks into a single make-based "style" command.
  * Removed redundant automated lint workflows to simplify CI.
  * Updated default build target to run the unified style checks.
  * No user-facing functionality changes.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -11,7 +11,7 @@ name: style
     branches:
       - master
 jobs:
-  make:
+  style:
     timeout-minutes: 15
     runs-on: ubuntu-24.04
     steps:


.github/workflows/style.yml
```diff
@@ -2,22 +2,20 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
-name: flake8
-'on':
+name: make
```
ivanovmg: I suppose the name of the pipeline should be something like `style` rather than `make`.

```suggestion
name: style
```

PR #962, opened 2025-10-16
Replace `ty` with mature configured `mypy`, fix typing issues

### LINKED ISSUE #896: `ty` is pre-release software and not ready for production use

deyna256:
Resolve #896

## Summary by CodeRabbit

* **Chores**
  * Added comprehensive type annotations, stricter static-analysis settings, reduced max line length to 100, and added a Makefile target for local type checks.
* **Chores / CI**
  * Removed the automated type-checking workflow from CI.
* **Bug Fixes**
  * Encoding detection now always returns decoded text (empty input → empty string) and a minor metric assignment bug was fixed.

### DIFF
--- a/aibolit/patterns/loop_outsider/loop_outsider.py
+++ b/aibolit/patterns/loop_outsider/loop_outsider.py
@@ -17,7 +17,7 @@ class LoopOutsider:
     def __init__(self) -> None:
         pass
 
-    def value(self, filename: str | os.PathLike) -> List[LineNumber]:
+    def value(self, filename: str | os.PathLike[str]) -> List[LineNumber]:
         """
         Returns the line number of loop outsiders found in file.
         """


literally-bug-creator:
@yegor256 @ivanovmg wdyt?

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 ignore = W504,EO001,EO002,EO003,EO004,EO005,EO008,EO009,EO01,EO02
-max-line-length = 120
+max-line-length = 100
```
ivanovmg, line 3: I suppose that this change goes beyond the scope of the present PR.

Makefile
```diff
@@ -7,12 +7,14 @@
...
 .SECONDARY:
 SHELL := bash
-.PHONY: all clean requirements test it install xcop flake8 pylint sphinx mypy lint e2e build coverage
+.PHONY: all clean requirements test it install xcop flake8 type pylint sphinx mypy lint e2e build coverage
 
 all: requirements install test it lint xcop sphinx
 
 lint: flake8 pylint mypy
 
+type: mypy
```
ivanovmg, line 16: Is there a specific reason to create the alias? I mean, we could just have used `mypy` instead on line 10. Or, am I missing something?

aibolit/ast_framework/computed_fields_catalog/chained_fields.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2019-2025 Aibolit
 # SPDX-License-Identifier: MIT
 
-from typing import Union, Any, Callable
+from typing import Union, Any, Callable, List
```
ivanovmg, line 4: Not that it is that critical, but starting from Python 3.9 they allowed using just regular `list`, `tuple` instead of `List` and `Tuple`.
Also, a more modern way suggests avoiding `Union[X, Y]` in favor of `X | Y`.

aibolit/metrics/cognitiveC/cognitive_c.py
```diff
@@ -108,9 +108,9 @@ def _not_nested_complexity(self, node: ASTNode, nested_level: int, method_name:
 
     def _method_name(self, node: ASTNode) -> str:
         if node.node_type == ASTNodeType.METHOD_DECLARATION:
-            return node.name
+            return cast(str, node.name)
         if node.node_type == ASTNodeType.METHOD_INVOCATION:
             if node.qualifier:
-                return node.qualifier + '.' + node.member
-            return node.member
+                return cast(str, node.qualifier) + '.' + cast(str, node.member)
```
ivanovmg, line 114: This should probably be more readable, if passes the type checker.
```suggestion
                return cast(str, node.qualifier + '.' + node.member)
```

aibolit/model/model.py
```diff
@@ -111,7 +120,10 @@ def fit_regressor(self, X, y, features=None, display=False) -> None:
         if features:
             X = X[features]
 
-        self.features_conf = {'features_order': X.columns}
+        if hasattr(X, 'columns'):
```
ivanovmg, line 123: I guess that `columns` is the attribute of a pandas dataframe. If so, then the type of `X` should probably be `pd.DataFrame | NDArray`.

aibolit/model/model.py
```diff
@@ -111,7 +120,10 @@ def fit_regressor(self, X, y, features=None, display=False) -> None:
...
+            self.features_conf = {'features_order': list(X.columns)}
+        else:
+            self.features_conf = {'features_order': list(range(X.shape[1]))}
```
ivanovmg, line 126: I wonder if this new branch is handled by the tests. If not, then probably it may be a good idea to make `X` of type `pd.DataFrame` only and remove `else` branch at all, like in the original code.
Moreover, in line 112 (original) we call `X[features]`, which will fail for numpy arrays, but should be ok for pandas dataframe.

ivanovmg (commented):
@literally-bug-creator please have a look at my comments.

I noticed that there is a lot of `cast`ing, when it comes to resolving the node type, which is not perfect, but I don't know if there is a way around it.

PR #1192, opened 2026-06-10
Preserve original error context in __count_value

### LINKED ISSUE #1181: `__count_value` exception wrapper in `aibolit/__main__.py` discards important error information
The exception wrapper around the per-metric call in `__count_value` discards everything useful about the underlying failure. In `aibolit/__main__.py` lines 148-150 the code reads:

```python
except Exception:
    exc_type, exc_value, exc_tb = sys.exc_info()
    raise Exception(f"Can't count {acronym} metric: {str(type(exc_value))}")
```

Three things go wrong here.

The new `Exception` carries only `str(type(exc_value))`, which renders as `<class 'KeyError'>` or `<class 'AttributeError'>`. The actual message (the missing key, the missing attribute, the file path that failed) is dropped before it reaches the user or the log.

Because the wrapper does not use `raise ... from exc_value`, Python's exception-chaining mechanism is lost too: the resulting traceback hides the line in the metric pipeline that triggered the failure, so the operator has to bisect the codebase to find out which `make()` or `value()` call broke.

The wrapper also collapses every error class (including programming bugs like `AttributeError`) into a generic `Exception`, which makes the catch in `main()` at line 917 unable to distinguish a real defect from a per-file analysis failure.

A minimal fix is to preserve both the message and the chain:

```python
except Exception as exc:
    raise RuntimeError(f"Can't count {acronym} metric on {java_file}: {exc}") from exc
```

Adding the source file name to the message also makes the failure reproducible when one file in a batch trips the run.

somepatt:
## Summary
- preserve the original exception message when `__count_value` fails
- raise `RuntimeError` with exception chaining and the analyzed file path
- add a regression test covering the preserved message and `__cause__`

## Verification
- `python3 -m pytest test/recommend/test_recommend_pipeline.py -q`

Closes #1181 

## Release Notes

* **Bug Fixes**
  * Enhanced error reporting to include more detailed context, distinguishing between different item types.
  * Improved exception handling to preserve original error information while providing clearer, more informative error messages.

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -137,6 +137,7 @@ def __count_value(value_dict, input_params, code_lines_dict, java_file: str, is_
     :return: None, it has side-effect
     """
     acronym = value_dict['code']
+    entity = 'metric' if is_metric else 'pattern'
     try:
         ast = AST.build_from_javalang(build_ast(java_file))
         val = value_dict['make']().value(ast)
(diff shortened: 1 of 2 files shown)


somepatt:
@yegor256 could you please review this PR?

PR #1193, opened 2026-06-10
Preserve string literals in RemoveComments

### LINKED ISSUE #1185: `RemoveComments.remove_comments` unintentionally deletes Java string literals content
`aibolit/utils/utils.py:13-19` implements `RemoveComments.remove_comments` with two unanchored regexes — `/\*.*?\*/` for block comments and `//.*?\n` for line comments — applied directly to the full Java source string. The function makes no attempt to skip text inside string or character literals, so any `//` or `/*` sequence that lives inside a Java `"..."` literal is treated as a comment and silently deleted.

This corrupts the very inputs fed into `IndentationCounter.__file_to_tokens` at `aibolit/metrics/spaces/SpaceCounter.py:25-28`, the only consumer of `remove_comments`. A class with a field like `String url = "http://example.com/path";` ends up collapsed into `String url = "http:` and then concatenated with the next line, because the `//.*?\n` regex eats from the slashes through the end of the line. A literal like `"/* tag */"` is wiped from the source the same way. After this, every space-based indentation metric the file feeds is computed on a truncated, line-merged version of the program — the variance and max numbers no longer describe the real source.

Running this snippet against the master tree reproduces the corruption end to end:

```python
from aibolit.utils.utils import RemoveComments
print(RemoveComments.remove_comments(
    'class A {\n    String url = "http://example.com";\n    int x;\n}\n'
))
```

The output omits the `int x;` line entirely because everything from `//` to the next newline is consumed.

A correct fix replaces the two `re.sub` calls with a small state machine (or a tokenizer) that walks the source one character at a time and only enters comment mode while outside a string or character literal, accounting for escaped quotes. The change belongs in `aibolit/utils/utils.py` and needs no signature change.

somepatt:
## Summary
- replace regex-based comment stripping with a small scanner that respects Java string and character literals
- preserve line breaks while removing real comments so indentation metrics keep the original line structure
- add regression coverage for `http://...` and `"/* ... */"` literals and update spacing expectations accordingly

## Verification
- `python3 -m pytest test/utils/test_remove_comments.py -q`
- `python3 -m pytest test/metrics/spaces/TestSpaces.py -q`
- `python3 -m pytest test/utils/test_encoding_detector.py -q`

Closes #1185

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comment removal to correctly preserve comment-like content inside string and character literals and to more reliably strip real inline and block comments (including trimming comment-only line whitespace).

* **Tests**
  * Added unit tests covering preservation of literals and removal of comments.
  * Updated indentation metric test expectations to reflect corrected values.

### DIFF
--- a/aibolit/utils/utils.py
+++ b/aibolit/utils/utils.py
@@ -8,86 +8,90 @@ class RemoveComments:
         pass
 
     @staticmethod
+    def _trim_comment_only_indent(result):
+        while result and result[-1] not in '\n\r':
+            if result[-1] not in ' \t':
+                return
+            result.pop()
+
+    @staticmethod
+    def _consume_comment(symbol, next_symbol, result, state):
+        if state['mode'] == 'line_comment':
+            if symbol == '\n':
+                result.append(symbol)
+                state['mode'] = None
+            return 1
+        if state['mode'] == 'block_comment':
+            if symbol == '*' and next_symbol == '/':
+                state['mode'] = None
+                return 2
+            if symbol == '\n':
+                result.append(symbol)
+            return 1
+        return 0
+
+    @staticmethod
+    def _consume_literal(symbol, state):
+        if state['mode'] not in {'string', 'char'}:
+            return False
+        if state['escaped']:
+            state['escaped'] = False
+        elif symbol == '\\':
+            state['escaped'] = True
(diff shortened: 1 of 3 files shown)


somepatt:
@yegor256 could you please review this PR?

PR #1223, opened 2026-06-24
Ignore lambda casts in P5

### LINKED ISSUE #343: P5 Force Type Casting Finder is wrong
This is my code:

```java
    public void mapsToSameObjects() {
        final Iterable<Scalar<Integer>> list = new Solid<>(
            new org.cactoos.list.Mapped<>(
                i -> (Scalar<Integer>) () -> i,
                new IterableOf<>(1, -1, 0, 1)
            )
        );
        new Assertion<>(
            "Can't map only once",
            list.iterator().next(),
            new IsEqual<>(list.iterator().next())
        ).affirm();
    }
```
Aibolit complains about line no.4: "P5 Force Type Casting Finder". It seems that this code can't be refactored. This is how it has to look. Or I'm wrong?

somepatt:
## Summary
- ignore P5 findings for casts whose expression is a lambda
- add regression coverage for the cactoos-style lambda cast reported in the issue

## Problem
`ForceTypeCastingFinder` currently reports casts like `(Scalar<Integer>) () -> i` as force type casting. In this case the cast applies to a lambda expression and is used to adapt it to the target functional type, which matches the false positive reported in #343.

## Fix
Skip `CAST` nodes whose expression is a `LAMBDA_EXPRESSION`.

## Verification
- `python3 -B -m unittest test.patterns.force_type_casting_finder.test_force_type_casting_finder`
- `python3 -B -m flake8 --max-line-length=120 aibolit/patterns/force_type_casting_finder/force_type_casting_finder.py test/patterns/force_type_casting_finder/test_force_type_casting_finder.py`
- `TMPDIR=/Users/iliaprokofev/.tmp-codex-aibolit python3 -B -m pylint aibolit/patterns/force_type_casting_finder/force_type_casting_finder.py test/patterns/force_type_casting_finder/test_force_type_casting_finder.py`

Closes #343

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type-cast detection so casts used with lambda expressions are no longer reported.
  * Reduced false positives in pattern results while keeping existing cast detection behavior intact.
* **Tests**
  * Added coverage for lambda-based casts to confirm they are ignored by the matcher.

### DIFF
--- a/aibolit/patterns/force_type_casting_finder/force_type_casting_finder.py
+++ b/aibolit/patterns/force_type_casting_finder/force_type_casting_finder.py
@@ -8,4 +8,8 @@ from aibolit.ast_framework import AST, ASTNodeType
 
 class ForceTypeCastingFinder:
     def value(self, ast: AST) -> List[int]:
-        return [cast.expression.line for cast in ast.proxy_nodes(ASTNodeType.CAST)]
+        return [
+            cast.expression.line
+            for cast in ast.proxy_nodes(ASTNodeType.CAST)
+            if cast.expression.node_type != ASTNodeType.LAMBDA_EXPRESSION
+        ]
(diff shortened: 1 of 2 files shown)


somepatt:
@yegor256 please review this PR.
It fixes a reproducible false positive in P5 for lambda adaptation casts and includes a regression test.

PR #1232, opened 2026-06-25
Deduplicate class-level patterns after decomposition

### LINKED ISSUE #1217: Duplicated violation reports for class-level patterns due to class decomposition in Aibolit
## Description

When a Java class is decomposed into components by `decompose_java_class()` 
(used internally by `aibolit recommend`), every component AST still contains 
the original `CLASS_DECLARATION` node. As a result, any pattern that searches 
for `ASTNodeType.CLASS_DECLARATION` directly (instead of operating on a single, 
file-level pass) fires once **per component**, not once per class.

This produces many duplicate findings for what is really a single violation. 
Affected patterns include at least:

- `non_final_class` (P-code for "Non final class")
- `er_class` ("Prohibited class name")
- `many_primary_ctors`
- `implements_multi`

(any pattern whose `value()` iterates `ast.proxy_nodes(ASTNodeType.CLASS_DECLARATION)`)

## Steps to Reproduce

1. Create a small Java class with several independent fields/methods 
   (so the usage graph splits into multiple strongly-connected components), 
   and a name containing a forbidden word, e.g.:

```java
package com.test;

import java.util.HashMap;
import java.util.Map;

public class ConfigManager {
    private Map<String, String> data = new HashMap<>();
    private int counter = 0;
    private String name = "default";

    public void setValue(String key, String value) { data.put(key, value); counter++; }
    public String getValue(String key) { return data.get(key); }
    public int getCounter() { return counter; }
    public void incrementCounter() { counter++; }
    public String getName() { return name; }
    public void setName(String n) { this.name = n; }
    public void reset() { data.clear(); counter = 0; name = "default"; }
    public boolean hasKey(String key) { return data.containsKey(key); }
    public int size() { return data.size(); }
}
```

2. Run:

```bash
aibolit recommend --full --filenames ConfigManager.java
```

## Actual Result

The same finding for the same class (line 6) is reported 12 times:
ConfigManager.java[6]: Prohibited class name (P4: 6.40)

ConfigManager.java[6]: Prohibited class name (P4: 6.40)

ConfigManager.java[6]: Prohibited class name (P4: 6.40)

ConfigManager.java[6]: Prohibited class name (P4: 3.20)

... (12 lines total, all for the same class/line)
I confirmed the root cause directly:

```python
from aibolit.ast_framework import AST, ASTNodeType
from aibolit.ast_framework.java_class_decomposition import decompose_java_class
from aibolit.utils.ast_builder import build_ast
from aibolit.patterns.er_class.er_class import ErClass

ast = AST.build_from_javalang(build_ast('ConfigManager.java'))
class_ast = ast.subtree(ast.root().types[0])
components = list(decompose_java_class(class_ast, 'strong'))
print(len(components))  # 12
print(sum(1 for c in components if ErClass().value(c)))  # 12 -- fires in every component
```

The class is split into 12 strongly-connected components, and `ErClass.value()` 
(like the other class-level patterns) re-detects the same `CLASS_DECLARATION` 
node in every single component.

## Expected Result

A class-level violation (wrong class name, non-final class, multiple primary 
constructors, multiple interfaces implemented) should be reported once per 
class, not once per decomposed component. This also distorts the "importance" 
score shown in the report, since the same line contributes multiple times 
with different importance values.

## Possible Fix

Either:
1. Deduplicate findings by `(pattern_code, line)` before adding them to 
   `code_lines_dict` / `results`, or
2. Skip class-level patterns during the per-component loop in 
   `calculate_patterns_and_metrics_with_decomposition` and run them once 
   on the whole class AST instead, or
3. Have `decompose_java_class` strip the class-level node info that isn't 
   relevant to the component, so class-level patterns naturally only match 
   on a single full-class pass.

I'm happy to submit a PR for this if a maintainer can confirm the preferred 
approach (1, 2, or 3 above, or another direction).

somepatt:
## Summary
- count class-level patterns only once per class during decomposition
- keep component-specific patterns and metrics unchanged
- add a regression test for duplicated `P4`/`P24` findings

## Problem
`calculate_patterns_and_metrics_with_decomposition()` runs every pattern on every decomposed component. Patterns that inspect `CLASS_DECLARATION` therefore report the same class once per component, which duplicates findings and distorts ranking.

## Fix
Precompute class-level pattern results on the whole class AST and attach them only to the first decomposed component. Other components keep an empty result for those patterns.

## Verification
- `python3 -B -m pytest test/recommend/test_recommend_pipeline.py -q`
- `python3 -B -m flake8 --max-line-length=120 aibolit/__main__.py test/recommend/test_recommend_pipeline.py`

Closes #1217

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decomposed recommendation analysis so class-level patterns are pre-filtered and evaluated only once across related components.
  * Prevented duplicate class-level matches in later components, ensuring consistent pattern counts and line details.
  * Suppressed class-level patterns are now correctly skipped during decomposed analysis without impacting other results.
* **Tests**
  * Added regression coverage confirming class-level pattern counts are reported exactly once and line details appear only for the intended component.
  * Added a test ensuring suppressed class-level patterns are not computed.

### DIFF
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -39,6 +39,7 @@ from aibolit.ml_pipeline.ml_pipeline import train_process, collect_dataset
 from aibolit.utils.ast_builder import build_ast
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
+CLASS_LEVEL_PATTERN_CODES = frozenset({'P4', 'P7', 'P9', 'P24'})
 
 
 def list_dir(path, files):
@@ -292,6 +293,11 @@ def calculate_patterns_and_metrics_with_decomposition(
             if node.node_type == ASTNodeType.CLASS_DECLARATION
         ]
         for class_ast in classes_ast:
+            class_level_pattern_results = {
+                pattern_info['code']: pattern_info['make']().value(class_ast)
+                for pattern_info in patterns_info
+                if pattern_info['code'] in CLASS_LEVEL_PATTERN_CODES
+            }
             for index, component_ast in enumerate(decompose_java_class(class_ast, 'strong')):
                 result_for_component: Dict[Any, Any] = {}
                 code_lines_dict: Dict[Any, Any] = OrderedDict()
@@ -304,8 +310,13 @@ def calculate_patterns_and_metrics_with_decomposition(
                         input_params[pattern_info['code']] = 0
                         code_lines_dict['lines_' + pattern_info['code']] = []
                     else:
-                        pattern = pattern_info['make']()
-                        pattern_result = pattern.value(component_ast)
+                        if pattern_info['code'] in CLASS_LEVEL_PATTERN_CODES:
+                            pattern_result = class_level_pattern_results[pattern_info['code']]
+                            if index > 0:
+                                pattern_result = []
+                        else:
+                            pattern = pattern_info['make']()
+                            pattern_result = pattern.value(component_ast)
                         input_params[pattern_info['code']] = len(pattern_result)
                         code_lines_dict['lines_' + pattern_info['code']] = pattern_result
 
--- a/aibolit/__main__.py
+++ b/aibolit/__main__.py
@@ -304,8 +304,6 @@ def calculate_patterns_and_metrics_with_decomposition(
                 input_params = OrderedDict()  # type: ignore
 
                 for pattern_info in patterns_info:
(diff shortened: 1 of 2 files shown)


somepatt:
@yegor256 please review this PR.

somepatt:
@yegor256 please review this PR.

somepatt:
@yegor256 please review this PR.

PR #1275, opened 2026-07-14
fix(#1274): detect actual getters and setters

### LINKED ISSUE #1274: LCOM4 treats every get/set-prefixed method as an accessor
## Problem

`Filters.filter_getters_setters()` removes every method whose name starts
with `get` or `set`, regardless of the method body.

As a result, ordinary methods such as `setup()` or `getaway()` may be
incorrectly excluded from the cohesion graph used by the LCOM4 metric.

There is already a comment in `aibolit/utils/filter.py`:

```python
# To-Fix: implement get/set detection with .body
```

The project also has `ClassicGetter` and `ClassicSetter` patterns that
identify accessors by inspecting their bodies, rather than relying only
on their names.

## Reproducer

```java
class FalseGetterSetter {

    private int state;

    void setup() {
        this.state++;
    }

    void run() {
        this.setup();
    }
}
```

`setup()` is not a setter, but it is currently removed because its name
starts with `set`.

Expected LCOM4:

```text
1
```

Actual LCOM4:

```text
2
```

## Expected behavior

Only actual getter and setter methods should be excluded. Methods that
merely have names beginning with `get` or `set` should remain in the
cohesion graph.

Mahan1341:
Closes #1274.

## Changes

- detect getters and setters by their method bodies instead of name prefixes only;
- keep ordinary methods such as `setup()` and `getaway()` in the LCOM4 cohesion graph;
- add a regression test covering false `get`/`set` prefix matches.

## Tests

- `uv run --frozen pytest test/metrics/lcom4/test_lcom4.py -q` — 10 passed;
- `flake8` — passed;
- `pylint` — 10.00/10;
- `mypy` — passed;
- `ruff check` — passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved getter/setter recognition used by the LCOM4 metric.
  * Methods whose names start with “get” or “set” are no longer misclassified; detection now uses method structure/behavior to determine true accessors.

* **Tests**
  * Added a new Java fixture with misleading “get/set” methods.
  * Added a unit test asserting the LCOM4 value for the new fixture.

### DIFF
--- a/aibolit/utils/filter.py
+++ b/aibolit/utils/filter.py
@@ -14,6 +14,7 @@ from javalang.tree import (
     Node,
     ReturnStatement,
     This,
+    StatementExpression,
 )
 
 FldExh = Tuple[str, Tuple[str, str]]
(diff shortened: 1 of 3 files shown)


Mahan1341:
@yegor256, could you please review this PR? I reproduced the issue with a regression test, fixed the getter/setter detection, addressed the review comment, and all local checks pass. Thank you.

PR #1327, opened 2026-07-24
feat (#1300): Add P34 pattern for returning empty string literals ()

### LINKED ISSUE #1300: Feature request: Add P35 pattern for returning empty string literals ("")
In the Aibolit White Paper (Appendix), pattern **P16 (Return null)** is defined as an anti-pattern because it forces developers to write defensive checks, increasing Cognitive Complexity. 

Returning an empty string `""` often acts as a "hidden null" and causes the exact same maintainability issues:
1. Callers still have to write guard clauses (e.g., `if (!str.isEmpty()) { ... }`), which increases Cyclomatic/Cognitive complexity.
2. It can lead to silent logic failures downstream, just like `null` does.

Since Aibolit is an ML-based recommender system, adding `return ""` as a new feature will allow the model to statistically evaluate if empty string returns negatively impact the maintainability metric, similar to how it evaluates P16.

Arinapakalova:
Closes #1300

This PR introduces P33 (Return Empty String), a new pattern to detect methods returning empty string literals ("").

## Summary by CodeRabbit

- **New Features**
  - Added a Java analysis pattern that detects `return ""`, including when the empty string is returned from a ternary’s `true` and `false` branches.
  - Outputs the source line numbers for each match.
- **Documentation**
  - Updated pattern documentation with typo fixes and a new “Return Empty String” section.
- **Bug Fixes**
  - Simplified PMD `NcssCount` rule configuration by removing an unnecessary option.
- **Tests**
  - Added unit tests and Java fixtures covering empty-string returns, non-empty returns, and multiple scenarios.
- **Chores**
  - Added SPDX headers to the affected pattern and test modules.

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -600,7 +600,7 @@ class Book {
 
 ```java
 class Book {
-  puplic static void foo() {
+  public static void foo() {
     //something
   }
 }
@@ -720,7 +720,7 @@ class Foo {
 ```java
 class Foo {
   void foo() {
-    white (true) {
+    while (true) {
       for (;;) { // here
       }
     }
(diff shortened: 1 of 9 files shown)


Arinapakalova:
@coderabbitai review

PR #1348, opened 2026-07-29
fix(#1342): exclude generated _tmp directory from PMD target discovery

### LINKED ISSUE #1342: 03-calculate-metrics.py should not analyze its own _tmp directory
## Summary

After `./_tmp` is created, `scripts/03-calculate-metrics.py` treats it as an analysis target.

## Actual behavior

`collect_analysis_targets()` includes every top-level directory, so once `_tmp` exists it is added to the PMD input set and processed like project source code.

This can produce an extra target such as `_tmp_pmd_out.csv` and causes the script to analyze generated artifacts.

## Expected behavior

Generated working directories such as `_tmp` should be excluded from analysis target discovery.

## Why this is a bug

Temporary output directories are internal implementation details and should never be fed back into the analysis pipeline.

ImilB:
Closes #1342
## Why

Once `_tmp` is created, the metrics script may include it as a normal
top-level analysis target and feed generated artifacts back into PMD.

## What

- exclude `_tmp` from `collect_analysis_targets()`

## Result

The script analyzes only source inputs and no longer creates a synthetic
`_tmp` PMD target.

## How it was verified

- reused the PMD target collection regression coverage

### DIFF
--- a/scripts/03-calculate-metrics.py
+++ b/scripts/03-calculate-metrics.py
@@ -8,6 +8,21 @@ from pathlib import Path
 import pandas as pd
 
 DIR_TO_CREATE = 'target/03'
+TMP_DIR = Path('./_tmp')
+METRICS_COLUMNS = [
+    'filename',
+    'cyclo',
+    'cyclo_method_avg',
+    'cyclo_method_min',
+    'cyclo_method_max',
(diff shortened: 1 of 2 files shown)


ImilB:
@yegor256 awaiting approval

